PR #2, opened 2017-06-20
WIP: Add TLV support and general refactorings

Freeaqingme:
Hi,
I was looking for proxyprotocol support for Golang where I could encapsulate a listener like what is possible with https://github.com/armon/go-proxyproto , except I needed to use version 2 with TLV.

Therefore I was happy to find your repo, and took the liberty to extend it a bit here and there. It's still a WIP, though I'm curious what you think about the (albeit also still-in-progress) API.

Dolf
- Freeaqingme


protocol.go
```diff
@@ -0,0 +1,152 @@
+package proxyproto
```
Freeaqingme: Just realized I forgot to add any copyright notice, will at least add that it's derived from https://github.com/armon/go-proxyproto/blob/master/protocol.go

pires:
I like the idea of the connection wrapping even though it seems out-of-scope for this library. However, don't like the fact the API is broken with this change.

It's been some time since I last picked on this so I'll need to take a few hours to review properly.

jeffreydwalter:
@pires are you planning on merging these changes? I am also interested in support for v1 and v2 with the Listener interface.

pires:
I'd like to merge it, yes but, unfortunately, I haven't had the time to properly review it. The fact it breaks the existing API is also doesn't help.

pires:
Sorry for closing, it was accidental.

pires:
Would be cool if you could resume the TLV support based on current master branch 🙇

Freeaqingme:
Good job on #8 , was planning of reviewing it later, but you move fast :)

Right now I've got different priorities so can't tell when I'll revisit this (same goes for https://github.com/golang/go/issues/20956 ). Having said that, I do expect to be working on some HTTP projects later this year when I'll surely need something similar.

Though, if that takes too long, I by no way would take offence if someone else would continue this work.

pires:
Sorry, I should've waited indeed. I may be back to this PR and try and adapt your TLV work. Thank you **very much**!

dekimsey:
Apologies for the drive-by I was looking into the package for a Vault issue I'm curious about. But how might this compare against the PR on elastic's fork [here](https://github.com/elastic/go-proxyproto/pull/1)? It seems to have a similar goal of supporting TLVs in go-proxyproto.

Freeaqingme:
@dekimsey I'd ask the author of that PR. Chances are he's seen this PR as well, and he can explain why he thought one way to go about it would be better than another.

pires:
Hey @Freeaqingme! Thank you for your involvement so far in helping w/ this library. Is there anything from this PR that you think is missing from the current tip?

pires:
Thank you very much for your contribution even if it didn't get merged.

PR #4, opened 2019-01-07
Accommodate for padded payloads

mgouline:
Hi there,

First of all, great library, it got me out of a jam. Thanks for spending time on it.

There's an edge case that I found when using [AWS target groups for network load balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#proxy-protocol). Unlike the original HAProxy implementation, theirs pads the payload to 84 bytes (instead of the 12 that the library expects). They do it on purpose, as confirmed with their support, and it technically doesn't violate the version 2 spec.

As a result, I made it work by allowing longer lengths in the validation and then flushing the rest of the buffer until the indicated length, to ensure it doesn't interfere with the payload following it.

Tested on HAProxy and AWS. Any feedback welcome.

### DIFF
--- a/v2_test.go
+++ b/v2_test.go
@@ -4,11 +4,26 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/binary"
+	"reflect"
 	"testing"
 )
 
 var (
 	invalidRune = byte('\x99')
+
+	// Lengths to use in tests
+	lengthPadded = uint16(84)
+
+	lengthEmptyBytes = func() []byte {
+		a := make([]byte, 2)
+		binary.BigEndian.PutUint16(a, 0)
+		return a
+	}()
+	lengthPaddedBytes = func() []byte {
+		a := make([]byte, 2)
+		binary.BigEndian.PutUint16(a, lengthPadded)
+		return a
+	}()
 
 	// If life gives you lemons, make mojitos
 	portBytes = func() []byte {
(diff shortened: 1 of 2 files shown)


v2_test.go
```diff
@@ -73,14 +73,6 @@ var invalidParseV2Tests = []struct {
...
 		ErrInvalidLength,
 	},
-	{
-		newBufioReader(append(append(append(SIGV2, PROXY, TCPv4), lengthV6Bytes...), fixtureIPv6Address...)),
-		ErrInvalidLength,
-	},
-	{
-		newBufioReader(append(append(append(SIGV2, PROXY, TCPv6), lengthV6Bytes...), fixtureIPv4Address...)),
-		ErrInvalidLength,
-	},
```
pires: Why did you remove this? And why didn't you had tests to (reproduce and) validate your scenario?

mgouline: Removed these because they're testing for the invalid length that is higher than expected, which is what the `>=` now allows.

Good point, I'll create some tests for my scenario. Stay tuned.

mgouline:
OK, I've now put back the two test cases that you mentioned, instead changing the expected length in accordance with the `>=` condition and added a new test for the padded payload use case that I implemented.

CI seems to be failing because of a missing goveralls token, however the unit tests are all passing.

pires:
Please, rebase master and squash your commits. Tests should run now.

mgouline:
All done, rebased master and squashed commits. Same problem with CI unfortunately:

```
The command "$HOME/gopath/bin/goveralls -coverprofile=coverage.out -service=travis-ci -repotoken $COVERALLS_TOKEN" exited with 2.
```

Could it be because the code is in my fork `mgouline/go-proxyproto` and Travis is expecting that token to be associated with `pires/go-proxyproto`?

pires:
Bummer, you were right. I'll have to look into this some time in the future. Thank you very much!

Issue #3, opened 2018-02-26
An example for a Dialer and Listener is required.


elico:
This is a really good library but it's missing examples for a Dialer and a Listener.
IE client and a server service.
I might write one and I was thinking about a simple "echo" client and server.

The server would accept a PROXY TCP protocol connection and will echo back(and log to stdout) if possible the request protocol(TCP4\TCP6 only), src and destination IP:PORT.

The client would open a connection to a PROXY enabled server(Listener) with a specific src and destination IP:PORT (TCP4\TCP6) and will receive the into the socket\connection a line of readable text containing the received IP:PORT.

I have seen an implementation of this but I forgot to save the links so I cannot copy and paste it and I will need to use some CPU to write it.
Any recommendations on where to start?

I have seen this implementation but it's still not sitting well for me
- https://github.com/Freeaqingme/go-proxyproto/blob/master/protocol.go

Other examples:
- https://github.com/armon/go-proxyproto
- https://github.com/kennylevinsen/serve2d
- https://github.com/nabeken/mikoi/blob/master/mikoi.go
- https://github.com/ninchat/go-proxyprotocol
- https://github.com/inkel/viaproxy/blob/master/conn.go
- https://github.com/djmaze/proxy-responder/blob/master/main.go

I wrote a TPROXY Server and Library in golang at:
- https://github.com/elico/go-linux-tproxy

Which has a basic TCP proxy server function that can eventually utilize the PROXY protocol in couple environments.

elico:
@pires couple examples and utilities at:

- http://gogs.ngtech.co.il/elicro/proxy-protocol-examples

I will try to add there another example which uses TPROXY later on

anisse:
Hi @pires ! I've published my listener here: https://github.com/anisse/proxylistener/blob/master/proxylistener.go ; I'd really like it to be used as an example in this library, so I was wondering how to proceed ?
 - should it need a full program that shows how to use it or is the README sufficient ?
 - should it go into the examples/ directory ? What if we want people to use it directly ? Maybe contrib/listener ?
 - should it be tested ? How ? With a server counterpart ?

What am I missing?

elico:
@anisse I have just written two proxies from TCP to SCTP and from SCTP to TCP at:
- http://gogs.ngtech.co.il/NgTech-LTD/sctp-proxies

I believe that a GoLang programmer should have the fundamentals to understand what a listener and a dialer is but a good and practical example always helps these who doesn't fully grasp the subject.
[...]

PR #8, opened 2019-06-15
api: introduce net.Listener and net.Conn wrappers

pires:
This is an attempt to partially address #3 with pieces of #2. The reason why I'm not merging #2 ~(or the code I have here)~ is because I don't feel confident enough it isn't introducing breaking changes to any existing users of this library who may be pulling from master branch.

@Freeaqingme @mgouline @elico @anisse can you PTAL?

**TODO**
- [x] fix `golint` output
- [x] add test(s) for any added functions wherever makes sense
- [ ] ~add `net.Dial` example in order to fix #2~

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,46 @@ $ go get -u github.com/pires/go-proxyproto
 
 ### Client (TODO)
 
-### Server (TODO)
+### Server
+
+```go
+package main
+
+import (
+	"log"
+	"net"
+	
+	proxyproto "github.com/pires/go-proxyproto"
+)
+
+func main() {
+	// Create a listener
+	addr := "localhost:9876"
+	list, err := net.Listen("tcp", addr)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


anisse:
LGTM, especially with the updated comments from the later commits.

It seems this implementation is better than my own wrapper, which is a bit more naive (https://github.com/anisse/proxylistener/blob/master/proxylistener.go) :
 - initially I really liked the idea of wrapping net.Listen() directly as I did, but I realize now it's much better to wrap a listener, which is an interface so it can be something else than what stdlib provides. I also allows passing a tuned tcplistener (with setdeadline for example)
 - In addition, it gives you better control with the additional read timeout (in addition to the ), which allows having different behaviour than just the accept timeout (which my implementation does not allow anyway)
 - I also didn't really like the use of sync.Once, but I see now that it makes sense in the context of having a Conn that can be used in the client and server case (mine only does the server case).

In conclusion, LGTM :+1:

PR #9, opened 2019-08-20
TLV capture and parsing for proxy protocol v2

stu-elastic:
* Adds `rawTLVs` to Header struct
 * API Change 
   - non-`io.EOF` errors encountered when reading from payload reader are returned
     from `parseVersion2` and `Read`
   - Padding is saved in `rawTLV`
 * TLV splitting is implemented by `Header.TLVs()`, TLV length validation without
   this explicit call
 * Includes helpers for parsing AWS VPCEs and SSL information
 * Parsing of other types possible by consumers via the `PP2Type` constants using
   the `TLV` and `PP2SSL` exported types

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+# DO NOT USE - Temporary Fork #
+Wait until https://github.com/pires/go-proxyproto/pull/9 is merged.  This fork will be deleted.
+
 # go-proxyproto
 
 [![Build Status](https://travis-ci.org/pires/go-proxyproto.svg?branch=master)](https://travis-ci.org/pires/go-proxyproto)
--- a/header.go
+++ b/header.go
@@ -39,6 +39,7 @@ type Header struct {
 	DestinationAddress net.IP
 	SourcePort         uint16
 	DestinationPort    uint16
+	rawTLVs            []byte
 }
 
 // RemoteAddr returns the address of the remote endpoint of the connection.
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -1,3 +1,6 @@
+# DO NOT USE - Temporary Fork #
```
Freeaqingme: I understand why you put this comment note here, but ideally it would be gone before this PR was merged. Maybe use a separate branch downstream for this PR?

Freeaqingme:
Given the overlap with #2 I'll see if I can give this code a try later this week and see if it also accomplishes what I was trying to do with #2. Not that that should be the goal in itself, but I'd like to see if - if we start changing the API - can do so with only a single change rather than having to change it multiple times.

README.md
stu-elastic: Yeah makes sense, I'll reopen with a new branch.

stu-elastic:
> I'd like to see if - if we start changing the API - can do so with only a single change rather than having to change it multiple times.

This is a backwards compatible change, if there are non-bwc API changes that are required for #2 then I'd suggest rolling a new major version while letting this go into a minor version release.

stu-elastic:
Closed in favor of https://github.com/pires/go-proxyproto/pull/10

Freeaqingme:
> This is a backwards compatible change

Good point. Though you are extending the public interface a bit, so I'd like to see if it also suffices for my original requirements. Just to see if we can prevent a future BC break. (or at least know so in advance)

pires:
Maybe it's time to do releases and stop getting this from master. dep/go modules have come a long way.

PR #11, opened 2019-08-28
Add possibility to handle proxy info differently depending on upstream IP

mikey179:
This introduces the possibility to decide on each new incoming connection whether the proxy header should be used, ignored, or the connection be rejected. It also allows to reject the connection when a proxy header is required but not send at all.

Without such a possibility usage of this library is prone to security attacks enabling simple faking of client IP addresses by accepting any proxy protocol header regardless of whether the upstream source is allowed to send one or not. The need for this is derived from our use case, where we have an application which accepts connections both through a proxy and direct (there's a reason for that). If we would allow the proxy header for direct connections it would allow clients to pretend they are from another IP address which we don't want.

While `NewConn` is part of the public API I decided to change it nevertheless as no official release or version of this library exists yet. In case you want to keep it as is I'd introduce another function  `NewConnWithPolicy` instead where the caller can provide the policy, and fall back to the `USE` policy in `NewConn`.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -1,6 +1,10 @@
 package proxyproto
 
-import "net"
+import (
+	"fmt"
+	"net"
+	"strings"
+)
 
 // PolicyFunc can be used to decide whether to trust the PROXY info from
 // upstream. If set, the connecting address is passed in as an argument.
@@ -30,34 +34,75 @@ const (
 	REQUIRE
 )
 
+// WithPolicy adds given policy to a connection when passed as option to NewConn()
+func WithPolicy(p Policy) func(*Conn) {
+	return func(c *Conn) {
+		c.ProxyHeaderPolicy = p
+	}
+}
+
 // LaxWhiteListPolicy returns a PolicyFunc which decides whether the
 // upstream ip is allowed to send a proxy header based on a list of allowed
-// IP addresses. In case upstream IP is not in list the proxy header will
-// be ignored.
-func LaxWhiteListPolicy(allowed []string) PolicyFunc {
-	return whitelistPolicy(allowed, IGNORE)
+// IP addresses and IP ranges. In case upstream IP is not in list the proxy
+// header will be ignored. If one of the provided IP addresses or IP ranges
+// is invalid it will return an error instead of a PolicyFunc.
+func LaxWhiteListPolicy(allowed []string) (PolicyFunc, error) {
+	allowFrom, err := parse(allowed)
+	if err != nil {
+		return nil, err
+	}
+
+	return whitelistPolicy(allowFrom, IGNORE), nil
+}
+
+// MustLaxWhiteListPolicy returns a LaxWhiteListPolicy but will panic if one
+// of the provided IP addresses or IP ranges is invalid.
+func MustLaxWhiteListPolicy(allowed []string) PolicyFunc {
+	pfunc, err := LaxWhiteListPolicy(allowed)
+	if err != nil {
+		panic(err)
+	}
+
+	return pfunc
 }
 
 // StrictWhiteListPolicy returns a PolicyFunc which decides whether the
 // upstream ip is allowed to send a proxy header based on a list of allowed
-// IP addresses. In case upstream IP is not in list reading on the
-// connection will be refused on the first read. Please note: subsequent
+// IP addresses and IP ranges. In case upstream IP is not in list reading on
+// the connection will be refused on the first read. Please note: subsequent
 // reads do not error. It is the task of the code using the connection to
-// handle that case properly.
-func StrictWhiteListPolicy(allowed []string) PolicyFunc {
-	return whitelistPolicy(allowed, REJECT)
+// handle that case properly. If one of the provided IP addresses or IP
+// ranges is invalid it will return an error instead of a PolicyFunc.
+func StrictWhiteListPolicy(allowed []string) (PolicyFunc, error) {
+	allowFrom, err := parse(allowed)
+	if err != nil {
+		return nil, err
+	}
+
+	return whitelistPolicy(allowFrom, REJECT), nil
 }
 
-func whitelistPolicy(allowed []string, def Policy) PolicyFunc {
+// MustStrictWhiteListPolicy returns a StrictWhiteListPolicy but will panic
+// if one of the provided IP addresses or IP ranges is invalid.
+func MustStrictWhiteListPolicy(allowed []string) PolicyFunc {
+	pfunc, err := StrictWhiteListPolicy(allowed)
+	if err != nil {
+		panic(err)
+	}
+
+	return pfunc
+}
+
+func whitelistPolicy(allowed []func(net.IP) bool, def Policy) PolicyFunc {
 	return func(upstream net.Addr) (Policy, error) {
-		upstreamIP, _, err := net.SplitHostPort(upstream.String())
+		upstreamIP, err := ipFromAddr(upstream)
 		if err != nil {
 			// something is wrong with the source IP, better reject the connection
 			return REJECT, err
 		}
 
 		for _, allowFrom := range allowed {
-			if allowFrom == upstreamIP {
+			if allowFrom(upstreamIP) {
 				return USE, nil
 			}
 		}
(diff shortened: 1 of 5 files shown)


pires:

pires:
I like this idea a lot! Can you please rebase and fix conflicts?

I'm totally fine with merging and breaking the API since there's no release and people can simply rely on the commit of their choosing.
Actually, I just removed the timeout parameter since it wasn't even being used, so yeah.

mikey179:
Rebased and fixed conflicts.

pires:
@mikey179 thank you very much! Sorry for the delay in coming back to you. Meanwhile, it seems folks from Hashicorp are interested in the functionality you are introducing (#16). Would be great to have both of you agreeing on the functionality and subsequent implementation before we merge. WDYT?

policy.go
```diff
@@ -0,0 +1,67 @@
...
+
+import "net"
+
+// PolicyFunc can be used to decide whether to trust the PROXY info from
+// upstream. If set, the connecting address is passed in as an argument.
+//
+// See below for the different policies.
+//
+// In case an error is returned the connection is denied.
+type PolicyFunc func(upstream net.Addr) (Policy, error)
```
jefferai: I think it would be good to add the type of incoming connection to the parameters. This allows you to change your response if the incoming is UNKNOWN, or to reject/accept v4/v6 without having to parse the address.

jefferai: Or alternately provide the actual proxy string as well as the net-derived Addr.

jefferai: I'm having no luck finding the issue that got reported to us with armon/go-proxyproto and a function in there similar to this, but I'm adding a comment here to say that I'm trying to find it because I believe the way for us to address it would have been to break the SourceChecker API in there, and I'm not sure whether the slightly different API here works around it. I'll keep digging and get back to you on that.

pires: To my knowledge - at least I didn't - ask for this. The author of the PR came up with the idea themselves. Feature wise, seems very similar what both of you are trying to achieve and I'm more than happy to include whatever you agree is the best option here.

jefferai: Yeah this is just a note to say that I might have more stuff to comment on here so please don't commit/merge this _quite_ yet. I just have to try to figure out where/what the report was. I'm striking out on both GitHub issues and customer issues, so far. Likely I'm just not getting keywords right.

pires: Sure, let's wait for the PR author to come back to you.

protocol.go
```diff
@@ -47,10 +61,11 @@ func (p *Listener) Addr() net.Addr {
 
 // NewConn is used to wrap a net.Conn that may be speaking
 // the proxy protocol into a proxyproto.Conn
-func NewConn(conn net.Conn) *Conn {
+func NewConn(conn net.Conn, proxyHeaderPolicy Policy) *Conn {
```
jefferai: Rather than this change, I would suggest passing in something like a `opts *ConnOpts`. That way if you need to be able to pass more things in the future you don't have to break API again.

protocol.go
```diff
@@ -118,13 +133,29 @@ func (p *Conn) SetWriteDeadline(t time.Time) error {
 	return p.conn.SetWriteDeadline(t)
 }
 
-func (p *Conn) readHeader() (err error) {
-	p.header, err = Read(p.bufReader)
+func (p *Conn) readHeader() error {
```
jefferai: Note that this function is called in a `once` several places -- Read, LocalAddr, RemoteAddr. If LocalAddr or RemoteAddr are called before Read, the error is ignored, which means you won't actually close the connection.

Those functions don't return an error, but probably calling them with an error value should set a flag in the conn object that tells read (or write) to exit with an error if/when they are called.

protocol.go
```diff
@@ -32,7 +34,19 @@ func (p *Listener) Accept() (net.Conn, error) {
 	if err != nil {
 		return nil, err
 	}
-	return NewConn(conn), nil
+
+	proxyHeaderPolicy := USE
+	if p.Policy != nil {
+		proxyHeaderPolicy, err = p.Policy(conn.RemoteAddr())
+		if err != nil {
```
jefferai: I think this is where the issue I mentioned before is. IIRC, what we found -- using the other lib and returning an error from the checking function -- is that sometimes returning an error from Accept that wasn't a net.Error would cause a listener to simply completely shut down rather than just close the conn.

Thing is -- I can't reproduce this behavior against this code. I even tried old versions of Go, in case it was an issue fixed somewhere. So at the moment it seems okay, and maybe it was related to some specific behavior of some specific wrapping listener.

policy.go
jefferai: See other comment; I think this is resolved.

mikey179: I‘d go for passing the parsed header, not the proxy string. The decision about the policy has to be deferred to the first read then, and can‘t be done on accept. Which means any error from the actual ˋPolicyFuncˋ is only returned on read (or accessing one of the addresses, which is still incorrect as you pointed out below). Not sure if this is problematic from the user perspective, as it might be unexpected to receive such an error at that point after the user already received a connection?

protocol.go
mikey179: I‘m not against such a change, but should it be really part of this PR? I think it would just inflate what this one is about and rather should be a separate one.

protocol.go
mikey179: You are right, the error shouldn’t get lost when a user asks for the addresses first before a read. Will update the PR.

protocol.go
jefferai: You're changing the API already; I'm suggesting the API change be more forward looking. I think it would certainly be in-scope although I defer to @pires for preference.

policy.go
jefferai: That's fair, it's nicer not to accept if you know you're not going to accept it. But perhaps then there needs to be a separate hook that allows you to control behavior based on the actual contents of the connection once it's accepted. v4/v6 is less of an issue but UNKNOWN is important -- if the proxy can't actually determine the originating address you may well want to drop the connection. (We added it to armon/go-proxyproto since a lot of users needed this behavior, so this is not a purely theoretical comment.)

protocol.go
pires: If we're changing the API already, I'd prefer to avoid changing it once more in a subsequent PR, particularly because this library is not tagged and folks most probably `go get` from master branch and there's the chance someone may get this functionality mid-flight and have the first version of the API instead of the final one - that I expect you to agree upon. Makes sense?

policy.go
pires: So what's the conclusion on this one?

protocol.go
pires: @mikey179 let us know if you can't find the time to do this. Maybe @jefferai could open a new PR that includes your commits from this branch and wrap it up? If none of you can do it, maybe I can. Thanks to both of you for your effort in collaborating 👏

protocol.go
pires: TBH I never heard of such behavior.

policy.go
mikey179: My current plan is to build the options variant mentioned below, and then see how passing a second hook for checking the proxy string looks like. It might be that I don‘t come up with a good design for that as I don’t have a usecase personally, but then  the API is still open to add this hook in another PR.

protocol.go
mikey179: Sorry, should have mentioned that I‘m in between two jobs this month (starting a new one in February) and using the time to travel a lot, not having access to a development machine. I hope to pick this up in February again.

pires: Awesome! Good luck in the new adventure.

mikey179: With commit a467388 the errors aren't ignored any longer, calling `LocalAddr()` or `RemoteAddr()` before `Read()` will store the error on the connection and return it on any subsequent read.

protocol.go
mikey179: Changed `NewConn()` with 001e22a to accept an arbitrary amount of options which can configure the connection.

policy.go
mikey179: Added an additional hook via 26a1d2a. In case the policy has been decided to `USE` the proxy header, a `Validator` can be used to inspect the parsed header. In case the validator returns an error this will mark the connection as erred and any subsequent read will return this error.

mikey179:
I think I addressed all concerns raised so far, but in case I missed anything just ping me.

policy.go
```diff
@@ -0,0 +1,74 @@
...
+	// a PROXY header is not present, subsequent reads do not. It is the task
+	// of the code using the connection to handle that case properly.
+	REQUIRE
+)
+
+// LaxWhiteListPolicy returns a PolicyFunc which decides whether the
+// upstream ip is allowed to send a proxy header based on a list of allowed
+// IP addresses. In case upstream IP is not in list the proxy header will
+// be ignored.
+func LaxWhiteListPolicy(allowed []string) PolicyFunc {
```
jefferai: Here and in the policy below my suggestion would be to allow CIDRs rather than IP addresses (where the CIDRs can be e.g. /32). It can make some workflows a bit easier because you can know that your LBs will all be in the same subnet but if they can come and go (e.g. if they're being switched out for upgrades) if new ones come up they will be automatically trusted.

It's not a big deal, but it is a nicety. Stdlib ParseCIDR unfortunately _requires_ a CIDR; an alternative could be https://godoc.org/github.com/hashicorp/go-sockaddr#NewIPAddr. There are also functions in there that can check whether a given address is contained in the parsed subnet.

jefferai:
Other than my suggestion around CIDRs I think this is looking good!

policy.go
mikey179: I don't like adding a dependency just for that use case, especially as this library doesn't have any other yet. I'd rather try to check if the string contains a slash, and then treat it as a CIDR using `net.ParseCIDR()`, and as IP address otherwise.

jefferai: Sounds fine, I did forget about IPNet.Contains in stdlib.

pires: I like the idea of supporting CIDRs, as well. Thank you, @mikey179!

mikey179: Added support for CIDRs in 3eee238. The policy creating functions now return an error in case an invalid CIDR or IP address is provided. In case you don't want to explicitly check the errors new Must* functions were added, so if your program reads this from a configuration you can be sure it panics if the configuration is wrong.

pires:
@jefferai can you please give it a last review?

jefferai:
I think it looks good!

PR #12, opened 2019-09-23
proxyv2 LOCAL commands still need to skip over dummy header fields.

jeremysf:
When receiving proxy v2 headers with the command field set to zero (i.e. LOCAL commands) the rest of the header should be skipped over until length.

From the spec:

> \x0 : LOCAL : the connection was established on purpose by the proxy without being relayed. The connection endpoints are the sender and the receiver. Such connections exist when the proxy sends health-checks to the server. The receiver must accept this connection as valid and must use the real connection endpoints and discard the protocol block including the family which is ignored.

and later in the spec:

> When a sender presents a LOCAL connection, it should not present any address so it sets this field to zero. Receivers MUST always consider this field to skip the appropriate number of bytes and must not assume zero is presented for LOCAL connections.

A real world example of the consequences: Amazon's Network Load Balancer (AWS NLB) sends health check packets with the LOCAL command, and uses the TCPv4 address family with dummy IP and port information. Not skipping over the header bytes following the LOCAL command up to length results in the extra header bytes passing through to the connection. In the case of HTTP health checks, this results in noise before the HTTP headers and subsequently, errors attempting to parse/server HTTP endpoints. 

Attached is an example packet capture of AWS NLB HTTP health checks (successfully parsed / responded to with this commit applied).

[AWS-NLB-HTTP-HealthCheck.zip](https://github.com/pires/go-proxyproto/files/3643425/AWS-NLB-HTTP-HealthCheck.zip)

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -90,7 +90,7 @@ func (p *Conn) Close() error {
 // the socket server.
 func (p *Conn) LocalAddr() net.Addr {
 	p.once.Do(func() { p.readHeader() })
-	if p.header == nil {
+	if p.header == nil || p.header.Command.IsLocal() {
 		return p.conn.LocalAddr()
 	}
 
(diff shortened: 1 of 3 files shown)


pires:

pires:
I've ran tests manually and it passes, so I'm merging.

PR #20, opened 2020-04-07
Add support for Azure extensions

igor-kupczynski:
@pires do you accept contributions to tlvparse extension package? If so I'd like to add Azure support, which is fairly similar to AWS.

The support is modeled on `aws.go` under `tlvparse`. The Azure spec is similar but
encodes `uint32` instead of a string. Because of that I've opted for an
additional boolean indicating if given value was found -- the zero value of
`uint32` is technically a valid header value.

Let me know what do you think.

BTW We currently have a similar implementation (diff func names, etc. but the logic is the same) in our codebase but we don't run it in production yet. I've tested it on a dev env with Azure Private Link set up, and the header is parsed correctly.

---

Azure spec: https://docs.microsoft.com/en-us/azure/private-link/private-link-service-overview#getting-connection-information-using-tcp-proxy-v2

### DIFF
--- a/tlvparse/azure.go
+++ b/tlvparse/azure.go
@@ -1,0 +1,48 @@
+// Azure's application extension to TLVs for Private Link Services
+// https://docs.microsoft.com/en-us/azure/private-link/private-link-service-overview#getting-connection-information-using-tcp-proxy-v2
+
+package tlvparse
+
+import (
+	"encoding/binary"
+
+	"github.com/pires/go-proxyproto"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


pires:
Hey @igor-kupczynski! Sure, I'm more than happy to include this as it seems correct based on the documentation you provided - I have no way to test this atm. Also, this is a no-brainer for me given the library public API is not changed and only people who care about this extension will actually import it.

Thank you very much!

igor-kupczynski:
Thanks!

PR #21, opened 2020-05-04
header: return err from reader.Peek()

igorwwwwwwwwwwwwwwwwwwww:
The current behaviour of `Read` is to discard the underlying error and return `ErrNoProxyProtocol`. That makes it tricky to debug what is wrong.

In my case, I was trying to nest `tls.NewListener` inside a `proxyproto.Listener` but got the ordering wrong, the underlying error was:

```
tls: first record does not look like a TLS handshake
```

This patch surfaces that error to the caller.

Note: It may be a good idea to do the same thing with the other `.Peek()` calls too.

### DIFF
--- a/header.go
+++ b/header.go
@@ -123,7 +123,12 @@ func (header *Header) TLVs() ([]TLV, error) {
 // Also, this operation will block until enough bytes are available for peeking.
 func Read(reader *bufio.Reader) (*Header, error) {
 	// In order to improve speed for small non-PROXYed packets, take a peek at the first byte alone.
-	if b1, err := reader.Peek(1); err == nil && (bytes.Equal(b1[:1], SIGV1[:1]) || bytes.Equal(b1[:1], SIGV2[:1])) {
+	b1, err := reader.Peek(1)
+	if err != nil {
+		return nil, err
+	}
+
+	if bytes.Equal(b1[:1], SIGV1[:1]) || bytes.Equal(b1[:1], SIGV2[:1]) {
 		if signature, err := reader.Peek(5); err == nil && bytes.Equal(signature[:5], SIGV1) {
 			return parseVersion1(reader)
 		} else if signature, err := reader.Peek(12); err == nil && bytes.Equal(signature[:12], SIGV2) {
--- a/header.go
+++ b/header.go
@@ -125,13 +125,32 @@ func Read(reader *bufio.Reader) (*Header, error) {
 	// In order to improve speed for small non-PROXYed packets, take a peek at the first byte alone.
 	b1, err := reader.Peek(1)
 	if err != nil {
+		if err == io.EOF {
+			return nil, ErrNoProxyProtocol
+		}
 		return nil, err
 	}
 
 	if bytes.Equal(b1[:1], SIGV1[:1]) || bytes.Equal(b1[:1], SIGV2[:1]) {
-		if signature, err := reader.Peek(5); err == nil && bytes.Equal(signature[:5], SIGV1) {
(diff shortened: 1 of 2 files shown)


pires:
Hey @igorwwwwwwwwwwwwwwwwwwww! Thanks for the report and patch. Could you add a test that mimics your scenario so we see the TLS error actually bubbling up? And since we're at it, could you also please change all other `Peek()` calls to return an error?

igorwwwwwwwwwwwwwwwwwwww:
@pires I've added a test with a reader that returns a custom error.

I also added some logic to handle EOF differently. Some tests are expecting EOF to return `ErrNoProxyProtocol`, and that seems reasonable.

The error handling got a bit gnarly, so if you have suggestions for how to refactor this, I'm happy to do so.

header.go
```diff
@@ -123,10 +123,34 @@ func (header *Header) TLVs() ([]TLV, error) {
...
 func Read(reader *bufio.Reader) (*Header, error) {
 	// In order to improve speed for small non-PROXYed packets, take a peek at the first byte alone.
-	if b1, err := reader.Peek(1); err == nil && (bytes.Equal(b1[:1], SIGV1[:1]) || bytes.Equal(b1[:1], SIGV2[:1])) {
-		if signature, err := reader.Peek(5); err == nil && bytes.Equal(signature[:5], SIGV1) {
+	b1, err := reader.Peek(1)
+	if err != nil {
+		if err == io.EOF {
+			return nil, ErrNoProxyProtocol
+		}
+		return nil, err
```
pires, line 131: IIUC this is the fix to the problem you were observing, right?

header.go
```diff
@@ -123,10 +123,34 @@ func (header *Header) TLVs() ([]TLV, error) {
...
+	}
+
+	if bytes.Equal(b1[:1], SIGV1[:1]) || bytes.Equal(b1[:1], SIGV2[:1]) {
+		signature, err := reader.Peek(5)
+		if err != nil {
+			if err == io.EOF {
+				return nil, ErrNoProxyProtocol
+			}
+			return nil, err
```
pires, line 140: What type of errors do you think may happen here that can't be wrapped w/ a `ErrNoProxyProtocol`?

header_test.go
```diff
@@ -41,6 +50,18 @@ func TestReadTimeoutV1Invalid(t *testing.T) {
...
+	var e errorReader
+	reader := bufio.NewReader(&e)
+	_, err := ReadTimeout(reader, 50*time.Millisecond)
+
+	if err == nil {
+		t.Fatalf("expected error %s", errReadIntentionallyBroken)
+	} else if err != errReadIntentionallyBroken {
+		t.Fatalf("expected error %s, actual %s", errReadIntentionallyBroken, err)
+	}
+}
```
pires, line 63: This is OK but I'd be happier w/ a test case that covers the real scenario you ran into, the TLS issue.

header.go
igorwwwwwwwwwwwwwwwwwwww: That is correct.

header.go
igorwwwwwwwwwwwwwwwwwwww: The underlying `Read` syscall could fail afaik. In practice this is unlikely to happen here because we're using bufio with a default buffer size of 4k. So I'd expect the first `Read` to have filled the buffer with at least 6 bytes. So this is mostly handling a theoretical case.

pires: Gotcha.

pires:
👏  https://github.com/pires/go-proxyproto/releases/tag/0.1.2

PR #22, opened 2020-05-07
integration test for misconfigured TLS server

igorwwwwwwwwwwwwwwwwwwww:
This is an integration test that reproduces a TLS server getting a non-TLS message.

Before https://github.com/pires/go-proxyproto/pull/21, this would return a non-descript `Proxy protocol signature not present` message.

With that patch, we now surface the underlying error, in this case `tls: first record does not look like a TLS handshake`.

It's quite verbose, mostly because the TLS stuff requires a lot of setup.

refs https://github.com/pires/go-proxyproto/issues/3

### DIFF
--- a/protocol_test.go
+++ b/protocol_test.go
@@ -6,6 +6,8 @@ package proxyproto
 
 import (
 	"bytes"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"net"
 	"testing"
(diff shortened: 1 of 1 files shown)


internal/testcert.go
```diff
@@ -0,0 +1,47 @@
...
+qzzVtxxr397vWrjrIgPbJpQvBsafG7b0dA4AFjwVbFLmQcj2PprIMmPcQrooz8vp
+jy4SHEg1AkEA/v13/5M47K9vCxmb8QeD/asydfsgS5TeuNi8DoUBEmiSJwma7FXY
+fFUtxuvL7XvjwjN5B30pNEbc6Iuyt7y4MQJBAIt21su4b3sjXNueLKH85Q+phy2U
+fQtuUE9txblTu14q3N7gHRZB4ZMhFYyDy8CKrN2cPg/Fvyt0Xlp/DoCzjA0CQQDU
+y2ptGsuSmgUtWj3NM9xuwYPm+Z/F84K6+ARYiZ6PYj013sovGKUFfYAqVXVlxtIX
+qyUBnu3X9ps8ZfjLZO7BAkEAlT4R5Yl6cGhaJQYZHOde3JEMhNRcVFMO8dJDaFeo
+f9Oeos0UUothgiDktdQHxdNEwLjQf7lJJBzV+5OtwswCWA==
+-----END RSA TESTING KEY-----`))
+
+func testingKey(s string) string { return strings.ReplaceAll(s, "TESTING KEY", "PRIVATE KEY") }
```
pires, line 47: I'm curious why this function is needed. Do you know?

pires: If it's really needed, and in order to make coveralls happy, adding a simple test to it would be great.

igorwwwwwwwwwwwwwwwwwwww: It was copied from the stdlib `src/net/http/internal/testcert.go` -- but I guess we can get rid of it, if that makes coveralls happy. :)

igorwwwwwwwwwwwwwwwwwwww:
@pires My guess is that it was to prevent automated leak detectors from sounding alarms, let's see what happens. :)

pires:
It seems everyone is happy. Or almost :D now that we know it works, why not move the two variables to the test file hence removing the introduction of a new package (`internal`)? Also, please meld your commits into one to keep history clean.

Thank you so very, very much for this and for your patience.

igorwwwwwwwwwwwwwwwwwwww:
@pires done!

PR #27, opened 2020-08-08
V1 improvement

navossoc:
Just two small changes.

- Fix protocol header for unknown protocols.

Before:
`PROXY UNKNOWN <nil> <nil> 0 0\r\n`

Now:
`PROXY UNKNOWN\r\n`

----------

- Set a larger capacity for the buffer to avoid extra allocations.

`So a 108-byte buffer is always enough to store all the line and a trailing zero
for string processing.`
http://www.haproxy.org/download/2.3/doc/proxy-protocol.txt

[]'s


v1.go
```diff
@@ -67,14 +67,16 @@ func parseVersion1(reader *bufio.Reader) (*Header, error) {
...
 	// As of version 1, only "TCP4" ( \x54 \x43 \x50 \x34 ) for TCP over IPv4,
 	// and "TCP6" ( \x54 \x43 \x50 \x36 ) for TCP over IPv6 are allowed.
-	proto := "UNKNOWN"
+	var proto string
 	if header.TransportProtocol == TCPv4 {
 		proto = "TCP4"
 	} else if header.TransportProtocol == TCPv6 {
 		proto = "TCP6"
+	} else {
+		return []byte("PROXY UNKNOWN\r\n"), nil
```
pires, line 76:
```suggestion
		// Unknown connection (short form)
		return []byte("PROXY UNKNOWN\r\n"), nil
```

navossoc: Sure

PR #30, opened 2020-09-07
Prefix error messages with package name

emersion:
This makes it easier to track down where errors come from.

### DIFF
--- a/header.go
+++ b/header.go
@@ -16,19 +16,19 @@ var (
 	SIGV1 = []byte{'\x50', '\x52', '\x4F', '\x58', '\x59'}
 	SIGV2 = []byte{'\x0D', '\x0A', '\x0D', '\x0A', '\x00', '\x0D', '\x0A', '\x51', '\x55', '\x49', '\x54', '\x0A'}
 
-	ErrCantReadProtocolVersionAndCommand    = errors.New("Can't read proxy protocol version and command")
-	ErrCantReadAddressFamilyAndProtocol     = errors.New("Can't read address family or protocol")
-	ErrCantReadLength                       = errors.New("Can't read length")
-	ErrCantResolveSourceUnixAddress         = errors.New("Can't resolve source Unix address")
-	ErrCantResolveDestinationUnixAddress    = errors.New("Can't resolve destination Unix address")
-	ErrNoProxyProtocol                      = errors.New("Proxy protocol signature not present")
-	ErrUnknownProxyProtocolVersion          = errors.New("Unknown proxy protocol version")
-	ErrUnsupportedProtocolVersionAndCommand = errors.New("Unsupported proxy protocol version and command")
-	ErrUnsupportedAddressFamilyAndProtocol  = errors.New("Unsupported address family and protocol")
-	ErrInvalidLength                        = errors.New("Invalid length")
-	ErrInvalidAddress                       = errors.New("Invalid address")
-	ErrInvalidPortNumber                    = errors.New("Invalid port number")
-	ErrSuperfluousProxyHeader               = errors.New("Upstream connection sent PROXY header but isn't allowed to send one")
+	ErrCantReadProtocolVersionAndCommand    = errors.New("proxyproto: can't read proxy protocol version and command")
+	ErrCantReadAddressFamilyAndProtocol     = errors.New("proxyproto: can't read address family or protocol")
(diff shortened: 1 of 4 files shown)


pires:
Thank you for the time you took to come up with this contribution. However, I never seen this pattern before and I don't think it makes sense. Also, this project only has two packages, `proxyproto` and `tlvparse`, the latter being totally optional for you, so it shouldn't be that hard to know from which package the error comes from.

pires (commented):
Thank you for the time you took to come up with this contribution. However, I never seen this pattern before and I don't think it makes sense. Also, this project only has two packages, proxyproto and tlvparse, the latter being totally optional for you, so it shouldn't be that hard to know from which package the error comes from.

emersion:
The confusion mainly comes from e.g. `net.Conn.Read` suddenly returning the error "Invalid address". Since I'm not calling explicitly a `proxyproto` function (the method is hidden behind an interface, and called far away from the `proxyproto.Listener` setup), figuring out what happens is complicated.

This is also a pattern used throughout the standard Go library.

pires:
Would you be so kind to link me an example in the Go std library so I can learn?

emersion:
A few examples found by grepping:

```
src/archive/tar/common.go
30:	ErrHeader          = errors.New("archive/tar: invalid tar header")
31:	ErrWriteTooLong    = errors.New("archive/tar: write too long")
32:	ErrFieldTooLong    = errors.New("archive/tar: header field too long")
33:	ErrWriteAfterClose = errors.New("archive/tar: write after close")
34:	errMissData        = errors.New("archive/tar: sparse file references non-existent data")
35:	errUnrefData       = errors.New("archive/tar: sparse file contains unreferenced data")
36:	errWriteHole       = errors.New("archive/tar: write non-NUL byte in sparse hole")
631:		return nil, errors.New("archive/tar: FileInfo is nil")
src/bytes/reader.go
54:		return 0, errors.New("bytes.Reader.ReadAt: negative offset")
80:		return errors.New("bytes.Reader.UnreadByte: at beginning of slice")
106:		return errors.New("bytes.Reader.UnreadRune: at beginning of slice")
109:		return errors.New("bytes.Reader.UnreadRune: previous operation was not ReadRune")
128:		return 0, errors.New("bytes.Reader.Seek: invalid whence")
131:		return 0, errors.New("bytes.Reader.Seek: negative position")
src/crypto/cipher/gcm.go
113:		return nil, errors.New("cipher: incorrect tag size given to GCM")
117:		return nil, errors.New("cipher: the nonce can't have zero length, or the security of the key will be immediately compromised")
125:		return nil, errors.New("cipher: NewGCM requires 128-bit block cipher")
195:var errOpen = errors.New("cipher: message authentication failed")
src/crypto/tls/conn.go
603:		return c.in.setErrorLocked(errors.New("tls: internal error: attempted to read record with pending application data"))
762:		return c.in.setErrorLocked(errors.New("tls: too many ignored records"))
1073:	errClosed   = errors.New("tls: use of closed connection")
1074:	errShutdown = errors.New("tls: protocol is shutdown")
1137:		return errors.New("tls: internal error: unexpected renegotiation")
1166:		return errors.New("tls: unknown Renegotiation value")
1194:		return c.in.setErrorLocked(errors.New("tls: too many non-advancing records"))
1316:var errEarlyCloseWrite = errors.New("tls: CloseWrite called before handshake complete")
1372:		c.handshakeErr = errors.New("tls: internal error: handshake should have had a result")
1429:		return errors.New("tls: VerifyHostname called on TLS server connection")
1432:		return errors.New("tls: handshake has not yet been performed")
1435:		return errors.New("tls: handshake did not verify certificate chain")
src/mime/mediatype.go
107:		return errors.New("mime: no media type")
113:		return errors.New("mime: expected slash after first token")
117:		return errors.New("mime: expected token after slash")
120:		return errors.New("mime: unexpected content after media subtype")
128:var ErrInvalidMediaParameter = errors.New("mime: invalid media parameter")
190:			return "", nil, errors.New("mime: duplicate parameter name")
```

… And so on…

pires:
I'm lazy! Thanks for the lesson learned. Will you want to do the same for the `tlvparse` package?

emersion:
Sure, will do!

emersion:
Actually, unless I'm mistaken, it doesn't seem like `tlvparse` creates any new error values. It just re-uses `proxyproto` errors.

pires (approved):

PR #31, opened 2020-09-07
Introduce HeaderProxyFromAddrs

emersion:
Clients usually need to create a Header they'll WriteTo the underlying
connection.

This commit allows users to easily create a Header from a source
and destination net.Addr. If the net.Addr cannot be represented
by proxyporoto, the transport information is left unspecified.

### DIFF
--- a/header.go
+++ b/header.go
@@ -43,6 +43,69 @@ type Header struct {
 	rawTLVs            []byte
 }
 
+// HeaderProxyFromAddrs creates a new PROXY header from a source and a
+// destination address. If version is zero, the latest protocol version is
+// used.
+//
+// The header is filled on a best-effort basis: if hints cannot be inferred
+// from the provided addresses, the header will be left unspecified.
+func HeaderProxyFromAddrs(version byte, sourceAddr, destAddr net.Addr) *Header {
+	if version < 1 || version > 2 {
+		version = 2
+	}
+	h := &Header{
+		Version:           version,
+		Command:           PROXY,
+		TransportProtocol: UNSPEC,
+	}
+	switch sourceAddr := sourceAddr.(type) {
+	case *net.TCPAddr:
+		destAddr, ok := destAddr.(*net.TCPAddr)
+		if !ok {
+			break
+		}
+		if len(sourceAddr.IP.To4()) == net.IPv4len {
+			h.TransportProtocol = TCPv4
+		} else if len(sourceAddr.IP) == net.IPv6len {
+			h.TransportProtocol = TCPv6
+		} else {
+			break
+		}
+		h.SourceAddress = sourceAddr.IP
+		h.DestinationAddress = destAddr.IP
+		h.SourcePort = uint16(sourceAddr.Port)
+		h.DestinationPort = uint16(destAddr.Port)
+	case *net.UDPAddr:
+		destAddr, ok := destAddr.(*net.UDPAddr)
+		if !ok {
+			break
+		}
+		if len(sourceAddr.IP.To4()) == net.IPv4len {
+			h.TransportProtocol = UDPv4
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -43,6 +43,61 @@ type Header struct {
 	rawTLVs            []byte
 }
 
+// HeaderProxyFromAddrs creates a new PROXY header from a source and a
+// destination address. The latest protocol version is used.
+func HeaderProxyFromAddrs(sourceAddr, destAddr net.Addr) *Header {
+	h := &Header{
+		Version: 2,
```
pires, line 50: This should be an optional argument that if unset then default to 2.

header.go
```diff
@@ -43,6 +43,61 @@ type Header struct {
...
+		destAddr, ok := destAddr.(*net.TCPAddr)
+		if !ok {
+			break
+		}
+		if len(sourceAddr.IP.To4()) == net.IPv4len {
+			h.TransportProtocol = TCPv4
+		} else if len(sourceAddr.IP) == net.IPv6len {
+			h.TransportProtocol = TCPv6
+		} else {
+			break
```
pires, line 64: Why not set as `UNSPEC`?

header.go
```diff
@@ -43,6 +43,61 @@ type Header struct {
...
+func HeaderProxyFromAddrs(sourceAddr, destAddr net.Addr) *Header {
+	h := &Header{
+		Version: 2,
+		Command: PROXY,
+	}
+	switch sourceAddr := sourceAddr.(type) {
+	case *net.TCPAddr:
+		destAddr, ok := destAddr.(*net.TCPAddr)
+		if !ok {
+			break
```
pires, line 57: Should we return an error instead?

header.go
emersion: `UNSPEC` should be the default value for `TransportProtocol`. I can make that explicit if you prefer.

header.go
emersion: I've chosen to never return an error and fill the fields in a best-effort fashion. In this case, that means setting `TransportProtocol` to `UNSPEC` if the function can't figure out what to do.

I wonder if it makes sense to return an error instead. What would the caller do with the error? Making the error fatal because we can't infer the right details wouldn't be great.

header.go
emersion: Makes sense, will change this.

header.go
pires: Makes it easier for people using this func.

header.go
pires: Well, I think this reasoning should be part of the function comment so consumer knows exactly what the contract is: a best-effort helper.

emersion: Good point, will expand the docs.

emersion:
All done!

header.go
```diff
@@ -43,6 +43,69 @@ type Header struct {
...
 }
 
+// HeaderProxyFromAddrs creates a new PROXY header from a source and a
+// destination address. If version is zero, the latest protocol version is
+// used.
+//
+// The header is filled on a best-effort basis: if hints cannot be inferred
+// from the provided addresses, the header will be left unspecified.
+func HeaderProxyFromAddrs(version byte, sourceAddr, destAddr net.Addr) *Header {
+	if version == 0 {
```
pires, line 53: Should we enforce `2` when `version < 1 && version > 2`?

header_test.go
```diff
@@ -294,3 +294,70 @@ func TestFormat(t *testing.T) {
 		}
 	}
 }
+
+func TestHeaderProxyFromAddrs(t *testing.T) {
```
pires, line 298: Would be great not to reduce coverage but I can survive it. Just letting you know.

emersion:
Done!

pires (approved):

Issue #28, opened 2020-09-07
Give constants a type


emersion:
Right now it's not easy to tell that UNSPEC/TCPv4/etc are `AddressFamilyAndProtocol`. Somethign like this would help:

```go
const (
	UNSPEC      AddressFamilyAndProtocol = '\x00'
	TCPv4       AddressFamilyAndProtocol = '\x11'
	…
)
```

pires:
Makes sense. Please, go ahead and open a PR! Glad to review.

PR #32, opened 2020-09-11
Make constants typed

emersion:
Change AddressFamilyAndProtocol, PP2Type and ProtocolCommandAndVersion
constants to specify an explicit type. This makes it clearer for users
to understand what constants can be used for a particular header field.
This also makes it so the constants are groupped near the type
definition in the docs.

Closes: https://github.com/pires/go-proxyproto/issues/28


pires (approved):

pires:

Issue #29, opened 2020-09-07
Stop exporting internal protocol details


emersion:
Some constants like `CRLF` and `SEPARATOR` shouldn't be exported, because they are internal details.

pires:
Makes sense. Please, go ahead and open a PR! Glad to review.

PR #33, opened 2020-09-11
Stop exporting CRLF and SEPARATOR

emersion:
These are internal V1 details, users of this package shouldn't need
them.

Closes: https://github.com/pires/go-proxyproto/issues/29

### DIFF
--- a/v1.go
+++ b/v1.go
@@ -9,8 +9,8 @@ import (
 )
 
 const (
-	CRLF      = "\r\n"
-	SEPARATOR = " "
+	crlf      = "\r\n"
+	separator = " "
 )
 
 func initVersion1() *Header {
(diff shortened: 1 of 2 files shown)


emersion:
Notes:

- I wonder if we should unexport `SplitTLVs` too. Is there a case where users would get raw TLVs and want to split them?
- The `SIGV1`/`SIGV2` variables are left exported because some users might want to manually recognize PROXY headers.

pires:
>     * I wonder if we should unexport `SplitTLVs` too. Is there a case where users would get raw TLVs and want to split them?

There are cases, yes. One example is https://github.com/pires/go-proxyproto/blob/master/tlvparse/ssl.go#L96. Users may have their own code that requires this.

>     * The `SIGV1`/`SIGV2` variables are left exported because some users might want to manually recognize PROXY headers.
Yep!

pires (approved):

emersion:
>Users may have their own code that requires this.

Ah, I see, makes sense then!

PR #36, opened 2020-09-11
Add Header.SetTLVs

emersion:
This is the counterpart for Header.TLVs: it populates the rawTLVs field in Header.

### DIFF
--- a/header.go
+++ b/header.go
@@ -112,6 +112,17 @@ func (header *Header) TLVs() ([]TLV, error) {
 	return SplitTLVs(header.rawTLVs)
 }
 
+// SetTLVs sets the TLVs stored in this header. This method replaces any
+// previous TLV.
+func (header *Header) SetTLVs(tlvs []TLV) error {
+	raw, err := JoinTLVs(tlvs)
+	if err != nil {
+		return err
+	}
+	header.rawTLVs = raw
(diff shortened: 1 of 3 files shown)


header.go
```diff
@@ -112,6 +112,17 @@ func (header *Header) TLVs() ([]TLV, error) {
 	return SplitTLVs(header.rawTLVs)
 }
 
+// SetTLVs sets the TLVs stored in this header. This method replaces any
+// previous TLV.
+func (header *Header) SetTLVs(tlvs []TLV) error {
```
pires, line 117: While `JoinTLVs` is tested, this actually isn't.

pires (commented):
Can you please share a scenario where this could be useful, maybe even contribute an example test?

mschneider82:
it would be nice to have this feature merged, I created a PR for traefik to support proxy protocol for tcp backend connections using this lib, with SetTLVs it would be possible to add TLS informations to v2 proxy header.

pires:
That's valuable info @mschneider82, thank you for that. I'm just waiting on @emersion to address the coverage issue. If they don't come back to me in a few days, I can pick it up.

emersion:
Thanks for completing this work, I was on leave for the last two weeks.

PR #37, opened 2020-10-01
Added example for Client to readme, fixes #3

### LINKED ISSUE #3: An example for a Dialer and Listener is required.

mschneider82:


pires (approved):
AWESOME! 👏

mschneider82:
@pires  you have to bump a new version tag, otherwise the example will not work (only on master), since the WriteTo function is new

pires:
It's fine, I think. Anyways, I'm expecting to tag 0.2.0 as soon as #36 lands ;)

pires:
`0.2.0` is tagged

mschneider82:
@pires go get still downloads v0.1.3, you missed the v in the version please tag it as v0.2.0 instead of 0.2.0

pires:
Damn!

pires:
Fixed!

PR #40, opened 2020-10-09
Update spec URL

emersion:
go-proxyproto implements features that appeared in newer revisions of
the spec. Link the latest one.

While at it, convert links to HTTPS.


pires (approved):
Thanks!

Issue #35, opened 2020-09-11
TLV.Length is redundant


emersion:
We already have `len(TLV.Value)`. Having an extra `Length` field allows for a potential mismatch between these two sources of truth. This isn't that of a big deal when parsing (because the library will set both correctly), however this is more of an issue when formatting a TLV (because the caller needs to set both correctly).

PR #42, opened 2020-10-15
Nuke TLV.Length

emersion:
len(tlv.Value) can be used instead, and reduces the chances of a
mismatch between the actual length and the user-supplied Length field
when the user generates TLVs.

Obviously, this is a breaking change.

Closes: https://github.com/pires/go-proxyproto/issues/35

### DIFF
--- a/header_test.go
+++ b/header_test.go
@@ -203,13 +203,12 @@ func TestSetTLVs(t *testing.T) {
 				DestinationPort:    2000,
 			},
 			tlvs: []TLV{{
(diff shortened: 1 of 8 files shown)


pires:
At a quick glance, this LGTM but can you please make sure coverage is not reduced?

emersion:
>can you please make sure coverage is not reduced?

I'm removing lines of code that were previously tested, so this is expected. I'm not sure what to do about it.

pires (approved):

Issue #43, opened 2020-10-17
Header.{Local,Remote}Addr always return a TCP address


emersion:
TCP is hard-coded regardless of the `TransportProtocol`: https://github.com/pires/go-proxyproto/blob/3e7e0f55e48bb58288f1716a88770db96bfc342d/header.go#L111

pires:
I'm wondering if we should fix this in a 0.2.x branch 🤔

pires:
Actually, this is on purpose as it's an implementation detail. The function returns the interface `net.Addr`.

emersion:
Yes, the API is fine. The implementation isn't though, since the v2 header can parse non-TCP addresses. So for instance `TransportProtocol` could be `UDP6` and `LocalAddr` would still return a `net.TCPAddr`.

pires:
This seems to be expected for V1 as discussed in #45. I'm open to change this for V2.

Issue #34, opened 2020-09-11
Unix socket addresses missing from Header


emersion:
`Header` only supports IP addresses. However the Unix socket header is defined as such:

```c
        struct {        /* for AF_UNIX sockets, len = 216 */
             uint8_t src_addr[108];
             uint8_t dst_addr[108];
        } unix_addr;
```

pires:
I totally overlooked this and, unfortunately, changing it at this point would result in a API breaking change. The only thing that can be done is to add two attributes specifically for this or break the API and release a major change to the lib. I'm divided. Suggestions?

emersion:
Maybe we can start with adding two fields, and try to come up with something better once we have more breaking changes in mind?

pires:
Guess we're all trying to avoid a major release of the lib that would make Go versioning complicated, so yeah, maybe.

emersion:
This library is still unstable (0.X), so releasing a new 0.3 version would allow for breaking changes anyways.

See https://blog.golang.org/publishing-go-modules:

> If you add to the public API, make a breaking change to a v0 module, or upgrade the minor or version of one of your dependencies, increment the MINOR version for your next release.

pires:
I'm divided. Let's introduce the breaking change to master and decide later what to do in terms of release.

pires:
Well, given #42 is a breaking change why not just go ahead and address this too? Setting milestone for 0.3.0.

PR #46, opened 2020-10-17
Change addresses to net.Addr

emersion:
The motivation is to allow Header to hold Unix socket addresses.

Instead of using an IP and a port, use net.Addr, which allows to
represent any network address.

The Header.{Local,Remote}Addr methods are removed since users can now
just access Header.{Source,Destination}Addr.

New helpers (TCPAddrs, UDPAddrs, UnixAddrs, IPs, Ports) allows users to
easily extract data from the generic net.Addr fields.

Closes: https://github.com/pires/go-proxyproto/issues/34
Closes: https://github.com/pires/go-proxyproto/issues/43

* * *

Tests haven't been updated yet, but I'd like to gather some feedback because this is a quite radical change. What do you think?


v2.go
```diff
@@ -107,19 +108,15 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
...
 			return nil, ErrInvalidAddress
 		}
-		header.SourceAddress = addr.Src[:]
-		header.DestinationAddress = addr.Dst[:]
-		header.SourcePort = addr.SrcPort
-		header.DestinationPort = addr.DstPort
+		header.SourceAddr = newIPAddr(header.TransportProtocol, addr.Src[:], addr.SrcPort)
+		header.DestinationAddr = newIPAddr(header.TransportProtocol, addr.Dst[:], addr.DstPort)
 	}
 	// TODO fully support Unix addresses
```
pires, line 121: Uncomment this block?

v2.go
```diff
@@ -157,38 +154,43 @@ func (header *Header) formatVersion2() ([]byte, error) {
...
 	buf.Write(addrDst)
 
-	portSrcBytes := func() []byte {
-		a := make([]byte, 2)
-		binary.BigEndian.PutUint16(a, header.SourcePort)
-		return a
-	}()
-	buf.Write(portSrcBytes)
+	if sourcePort, destPort, ok := header.Ports(); ok {
+		portBytes := make([]byte, 2)
```
pires, line 185: Shouldn't it be 4 instead?

pires (commented):
It's looking good to me so far.

v2.go
emersion: The part is an `uint16`, so 2 bytes, unless I'm missing something? `portBytes` is overwritten for each port (source, then dest).

I suppose we could instead put the source port in `portBytes[0:2]` and the destination port in `portBytes[2:4]`, then write the whole thing.

pires: Nevermind, I had misread the code and it's fine to overwrite as there's a write to the buffer before that.

emersion:
All done! This is ready for another review.

pires (changes requested):
Overall LGTM but we're missing changes to `README.md` to reflect the breaking changes.

Also, I'll need a bit more time to review this properly.

Thank you very much for the work you've put into this 🤩

emersion:
>we're missing changes to README.md to reflect the breaking changes

Fixed

>Also, I'll need a bit more time to review this properly.

Sure, take your time :)

>Thank you very much for the work you've put into this

Thanks for maintaining this library!

pires (approved):
🎉

PR #48, opened 2020-10-28
Add more robust protocol mismatch handling in Header.formatVersion1

### LINKED ISSUE #6: Mixing IPv6 and IPv4 in header
I noticed while doing a little testing that if you generate the SourceAddress and DestinationAddress, it's possible to get an IPv4 address and an IPv6 address.

Example:
```
net.LookupIP("localhost")
```

Can return `::1` as the net.IP, while it's up to the user to verify that they're sending IPv4/IPv6 addresses with v4/v6 protocols, I'm wondering if this `writeVersion1` and `writeVersion2` functions should return an error if the IPs return `nil` for the protocol being sent.

I'm willing to do the PR for it, but I just thought checking if the validation should be in the package or in the user code is the right way.

emersion:
And add a bunch of invalid tests.

Fixes #6 (for v1 as v2 is already covered).

### DIFF
--- a/header_test.go
+++ b/header_test.go
@@ -447,27 +447,90 @@ func TestFormat(t *testing.T) {
 	if _, err := validHeader.Format(); err != nil {
 		t.Fatalf("shouldn't have thrown error %q", err.Error())
 	}
+}
 
-	invalidHeader := &Header{
-		Version:           3,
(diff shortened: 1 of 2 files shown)


pires (approved):
Great job, thanks a lot!

PR #49, opened 2020-10-28
Allow IPv6 compat addresses when parsing TCPv6 in V1 header

emersion:
Addresses like 0:0:0:0:0:ffff:7f00:1 should be allowed when parsing
IPv6 protocols.

* * *

Not sure how to make this work, so submitting as a draft so that we're at least aware of the issue...

### DIFF
--- a/header_test.go
+++ b/header_test.go
@@ -13,19 +13,22 @@ import (
 // Stuff to be used in both versions tests.
 
 const (
-	NO_PROTOCOL  = "There is no spoon"
-	IP4_ADDR     = "127.0.0.1"
-	IP6_ADDR     = "::1"
-	PORT         = 65533
-	INVALID_PORT = 99999
+	NO_PROTOCOL     = "There is no spoon"
+	IP4_ADDR        = "127.0.0.1"
+	IP6_ADDR        = "::1"
+	IP6_COMPAT_ADDR = "0:0:0:0:0:ffff:7f00:1"
+	PORT            = 65533
+	INVALID_PORT    = 99999
 )
 
 var (
-	v4ip = net.ParseIP(IP4_ADDR).To4()
-	v6ip = net.ParseIP(IP6_ADDR).To16()
+	v4ip       = net.ParseIP(IP4_ADDR).To4()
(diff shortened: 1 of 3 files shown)


pires:
Wait, didn't #49 fix this?

pires:
Ah, this is about IP4to6 where an IPv4 address can be encoded in IPv6, as the example you show up above. I think this is a good-to-have and not blocking 0.3.0 on it.

emersion:
Yes, I agree

isedev:
How about logic along these lines?

```
func parseV1IPAddress(v6 bool, addr string) (ip net.IP, err error) {
	ip := net.ParseIP(addr)
	tryV4 := ip.To4()
	hasDot := (strings.Index(addr, ".") != -1)
	hasColon := (strings.Index(addr, ":") != -1)
	if (!v6 && tryV4 != nil && !(hasColon && hasDot)) || (v6 && !hasDot) {
		err = errors.New("no good")
	}
	return
}
```

Which give these results:
```
addr                 error with v6=false?     error with v6=true?
8912:adfb::0124      yes                      no
10.10.0.10           no                       yes
::ffff:0a0a:0a0a     yes                      no
::ffff:10.10.10.10   yes                      yes
```

The ParseIP() and To4() outputs:
```
addr                 ParseIP                                       To4
8912:adfb::0124      len=16  ip=8912adfb000000000000000000000124   len=0  ip=
10.10.0.10           len=16  ip=00000000000000000000ffff0a0a000a   len=4  ip=0a0a000a
::ffff:0a0a:0a0a     len=16  ip=00000000000000000000ffff0a0a0a0a   len=4  ip=0a0a0a0a
::ffff:10.10.10.10   len=16  ip=00000000000000000000ffff0a0a0a0a   len=4  ip=0a0a0a0a
```

The `::ffff:10.10.10.10` is legal IPv6 text but not supported by the proxy protocol standard, as I read it.

isedev:
On similar note, `HeaderProxyFromAddrs` will silently assume TCP4 if given a IP4to6 address as source.
Not convinced that's proper behaviour. Can't see how to change that though without also changing function signature...

PR #51, opened 2020-10-29
exporting underlying Connection

### LINKED ISSUE #50: Export underlying connection
Hi,

It would be good to export the underlying `proxyproto.Conn.conn` connection. 
Since proxyproto.Conn is not an `net.TCPConn`, and I need to cast it.

mschneider82:
this blocks implementing the listener in traefik

https://github.com/traefik/traefik/pull/7320

If you accept it, please create a new version tag, thank you!

closes #50

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -147,6 +147,24 @@ func (p *Conn) RemoteAddr() net.Addr {
 	return p.header.SourceAddr
 }
 
+// TCPConn returns is the underlying connection as TCPConn
+func (p *Conn) TCPConn() (conn *net.TCPConn, ok bool) {
+	conn, ok = p.conn.(*net.TCPConn)
+	return
+}
+
+// UnixConn returns is the underlying connection as UnixConn
+func (p *Conn) UnixConn() (conn *net.UnixConn, ok bool) {
+	conn, ok = p.conn.(*net.UnixConn)
+	return
+}
+
+// UDPConn returns is the underlying connection as UDPConn
+func (p *Conn) UDPConn() (conn *net.UDPConn, ok bool) {
+	conn, ok = p.conn.(*net.UDPConn)
+	return
+}
+
 // SetDeadline wraps original conn.SetDeadline
 func (p *Conn) SetDeadline(t time.Time) error {
 	return p.conn.SetDeadline(t)
--- a/protocol_test.go
+++ b/protocol_test.go
@@ -470,6 +470,46 @@ func TestReadingIsRefusedOnErrorWhenLocalAddrRequestedFirst(t *testing.T) {
 	}
 }
 
+func Test_ConnectionCasts(t *testing.T) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+
+	policyFunc := func(upstream net.Addr) (Policy, error) { return REQUIRE, nil }
+
+	pl := &Listener{Listener: l, Policy: policyFunc}
+
+	go func() {
+		conn, err := net.Dial("tcp", pl.Addr().String())
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
+		defer conn.Close()
+		conn.Write([]byte("ping"))
+	}()
+
+	conn, err := pl.Accept()
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+	defer conn.Close()
+
+	proxyprotoConn := conn.(*Conn)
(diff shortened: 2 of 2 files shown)


pires (commented):
I'm trying to understand the issue that drives this change so this feedback is preliminary and assumes the problem can only be addressed by exposing the underlying `net.Conn`.

From the top of my head, I think a better option here would be for the attribute not to be exported but for an accessor (getter) to be added to `Conn`, say
```go
func (p *Conn) Raw() net.Conn {
    return p.conn
}
```

mschneider82:
A getter is fine to me. Should I change my pr or will you add it to master? (Raw needs a comment)

pires:
Go ahead! You started the work, after all. Thank you!

pires:
Also, can you please add tests that validate casts to TCP and UDP?

mschneider82:
What do you think? Or change it to Raw() ?

pires:
I think this is good as we have other pieces of code that validate casts as well 👍 @emersion do you want to comment?

pires (approved):
LGTM! If no discussion arises, will merge over the weekend.

protocol.go
```diff
@@ -147,6 +147,24 @@ func (p *Conn) RemoteAddr() net.Addr {
 	return p.header.SourceAddr
 }
 
+// TCPConn returns is the underlying connection as TCPConn
+func (p *Conn) TCPConn() (conn *net.TCPConn, ok bool) {
```
emersion, line 151: Instead of exposing one helper per connection type, I'd prefer to expose an `Unwrap` function that returns the nested connection. This would work for `net.Conn` types that aren't from the stdlib, and wouldn't put more work on callers:

```go
conn, ok := conn.Unwrap().(*net.TCPConn)
```

(The name `Unwrap` comes from the similar mechanism used by the `errors` package.)

What do you think?

emersion: Ah, I see this was already suggested earlier: https://github.com/pires/go-proxyproto/pull/51#pullrequestreview-519867029

emersion:
>If you accept it, please create a new version tag, thank you!

FWIW, you can upgrade to a specific commit with a need for a new version:

```
go get github.com/pires/go-proxyproto@<commit>
```

protocol.go
pires: Thing here is we now have code like the following https://github.com/pires/go-proxyproto/blob/master/header.go#L96 which does not, in this case, return the nested source and destination `net.Addr` and puts the onus on the consumer to do the casting themselves as suggested above. Therefore, I am accepting this change as is.

Thank you for your feedback, @emersion.

pires:
Thank you very much for your contribution, @mschneider82!

protocol.go
emersion: The addresses coming from the header are unrelated to the underlying connection. I could build a proxy from TCP to WebSockets which forwards the TCP addresses via the PROXY protocol, for instance.

guysv:
Hey. say I called the proxy wrapper's `Read()`, then unwrapped it to get the underlying connection, then tried to read from that underlying connection. Aren't I going to miss data that might reside in the proxy wrapper's buffered reader's internal buffer?

emersion:
Indeed, that sounds correct. We should probably document that.

guysv:
Also, I guess we could provide a method to drain the internal buffer (however that is done) because someone will probably will be interested in that data.

emersion:
This isn't really possible, once some data has been read from the underlying `net.Conn`, there's no way to put it back as if it wasn't read.

guysv:
Oh yes. of course. I was talking about exposing api to retrieve the data, and let the caller figure out what to do with it. I did not mean to get the data back into the connection.

emersion:
Users can read the buffered data from the `proxyproto.Conn`. Or am I missing something?

guysv:
Well, without calling the internal reader's `Reader.Buffered` you can't tell how much you can read without triggering another read from the connection itself.

emersion:
Oh, so you mean that you want to get the underlying `bufio.Reader`?

guysv:
Just proxying `Reader.Buffered` will do too. I just wonder if wouldn't you guys prefer to have a unified method that will both drain any remaining data from the internal buffer, and return the underlying conn. I'd be happy to send a PR for both approaches.

emersion:
I'm worried about people asking to add methods for other `bufio.Reader` functions in the future.

The stdlib just [exposes the `bufio.Reader`](https://golang.org/pkg/net/textproto/#Reader) in these situations. I think it would be fine to do the same.

PR #52, opened 2020-11-07
Add Conn.ProxyHeader

emersion:
This function returns the full proxy protocol header, if available.
Sometimes the local/remote addresses aren't enough and the caller is
interested in e.g. TLVs.


pires (approved):
LGTM, thank you!

PR #55, opened 2020-11-25
Add Generic Underlying Connection Getter

guysv:
After the type-specific underlying connection getters were introduced in #51 There's still use for a generic getter.

For example: there might be more wrappers to the connection, and if all wrappers could implement the `Conn() net.Conn` interface, a trivial resolver could unwrap them all and get to the actual-underlying connection in a generic way.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -154,19 +154,36 @@ func (p *Conn) RemoteAddr() net.Addr {
 	return p.header.SourceAddr
 }
 
-// TCPConn returns is the underlying connection as TCPConn
+// Raw returns the underlying connection which can be casted to
+// a concrete type, allowing access to specialized functions.
+//
+// Use this ONLY if you know exactly what you are doing. 
+func (p *Conn) Raw() net.Conn {
+	return p.conn
+}
+
+// TCPConn returns the underlying TCP connection,
+// allowing access to specialized functions.
+//
+// Use this ONLY if you know exactly what you are doing. 
 func (p *Conn) TCPConn() (conn *net.TCPConn, ok bool) {
 	conn, ok = p.conn.(*net.TCPConn)
 	return
 }
 
-// UnixConn returns is the underlying connection as UnixConn
+// UnixConn returns the underlying Unix socket connection,
+// allowing access to specialized functions.
+//
+// Use this ONLY if you know exactly what you are doing. 
 func (p *Conn) UnixConn() (conn *net.UnixConn, ok bool) {
 	conn, ok = p.conn.(*net.UnixConn)
 	return
 }
 
-// UDPConn returns is the underlying connection as UDPConn
+// UDPConn returns the underlying UDP connection,
+// allowing access to specialized functions.
+//
+// Use this ONLY if you know exactly what you are doing. 
 func (p *Conn) UDPConn() (conn *net.UDPConn, ok bool) {
 	conn, ok = p.conn.(*net.UDPConn)
 	return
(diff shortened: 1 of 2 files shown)


emersion (approved):
LGTM!

protocol.go
```diff
@@ -154,6 +154,11 @@ func (p *Conn) RemoteAddr() net.Addr {
 	return p.header.SourceAddr
 }
 
+// Conn returns is the underlying connection
+func (p *Conn) Conn() net.Conn {
```
pires, line 158: `p.Conn.Conn()` is stuttering. IIRC I had suggested in some other past PR to have this named `Raw()`. I don't remember any disagreement. Would you be ok w/ change it?

Also, can you please document further (as function comment) what this function can be used for and any concerns about it, eg buffered data?

pires (commented):
Small change, LGTM but have a couple requests that are not blockers.

protocol.go
guysv: Done.

guysv:
Done with @pires requests, (bump, not how GH's PR etiquette is working here).

protocol.go
```diff
@@ -154,19 +154,52 @@ func (p *Conn) RemoteAddr() net.Addr {
...
 }
 
+// Raw returns is the underlying connection
+//
+// The returned connection can casted to its concrete type
+// and allow access to specialized functions.
+//
+// Note that reading from the returned connection will lose
+// sync with internal buffers used in the proxy wrapper, and
+// will likely produce wroung results.
```
pires, line 164:
```suggestion
// Raw returns the underlying connection which can be casted to
// a concrete type, allowing access to specialized functions.
//
// Use this ONLY if you know exactly what you are doing.

File: protocol.go [hunk 3]
```diff
@@ -154,19 +154,52 @@ func (p *Conn) RemoteAddr() net.Addr {
...
+}
+
 // TCPConn returns is the underlying connection as TCPConn
+//
+// This function can be used to access functions specific to the
+// underlying connection type.
+//
+// Note that reading from the returned connection will lose
+// sync with internal buffers used in the proxy wrapper, and
+// will likely produce wroung results.
```
pires, line 176:
```suggestion
// TCPConn returns the underlying TCP connection,
// allowing access to specialized functions.
//
// Use this ONLY if you know exactly what you are doing. 
```

protocol.go
```diff
@@ -154,19 +154,52 @@ func (p *Conn) RemoteAddr() net.Addr {
...
 }
 
 // UnixConn returns is the underlying connection as UnixConn
+//
+// This function can be used to access methods specific to the
+// underlying connection type.
+//
+// Note that reading from the returned connection will lose
+// sync with internal buffers used in the proxy wrapper, and
+// will likely produce wroung results.
```
pires, line 189:
```suggestion
// UnixConn returns the underlying Unix socket connection,
// allowing access to specialized functions.
//
// Use this ONLY if you know exactly what you are doing. 
```

protocol.go
```diff
@@ -154,19 +154,52 @@ func (p *Conn) RemoteAddr() net.Addr {
...
 }
 
 // UDPConn returns is the underlying connection as UDPConn
+//
+// This function can be used to access methods specific to the
+// underlying connection type.
+//
+// Note that reading from the returned connection will lose
+// sync with internal buffers used in the proxy wrapper, and
+// will likely produce wroung results.
```
pires, line 202:
```suggestion
// UDPConn returns the underlying UDP connection,
// allowing access to specialized functions.
//
// Use this ONLY if you know exactly what you are doing. 
```

pires (changes requested):
I've made the comments concise while fixing some typos.

guysv:
done. thanks for the typo fix ^_^

pires (approved):
Thank you very much for your contribution!

PR #57, opened 2020-12-09
Fixed incorrect truncated tlv error that prevents reading empty (0 byte) TLVs

astromechza:
We hit this issue when reading a 0 byte custom TLV value. Testing identified that this was because the length was 0 bytes and the length check in this PR was affecting it.

### DIFF
--- a/tlv.go
+++ b/tlv.go
@@ -55,7 +55,7 @@ func SplitTLVs(raw []byte) ([]TLV, error) {
 		tlv := TLV{
 			Type: PP2Type(raw[i]),
 		}
-		if len(raw)-i <= 3 {
+		if len(raw)-i <= 2 {
 			return nil, ErrTruncatedTLV
 		}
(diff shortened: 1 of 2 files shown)


emersion:
This change looks good. Can you add a test with a 0-length TLV?

astromechza:
@emersion I've added the test now and checked that it fails on the old code 👍 

Any idea how we get a new release cut?

astromechza:
@pires what do you think of this diff?

emersion (approved):
LGTM, thanks!

emersion:
>Any idea how we get a new release cut?

Once this PR is merged, you can just `go get github.com/pires/go-proxyproto@<commit>` in your project to get the fix. Or wait for @pires to tag a new release.

pires (approved):
This is a great catch, thanks a lot, @astromechza!

pires:
Tagged v0.3.3 and will handle release notes over the coming days.

PR #61, opened 2021-01-17
Support v1 UNKNOWN and v2 UNSPEC when command is LOCAL

### LINKED ISSUE #60: Parsing v2 UNSPEC and v1 PROXY UNKNOWN not implemented
Hi! I found that this library is missing the ability to parse these two kinds of header, which are required by the protocol and mandatory to implement, while trying to resolve XTLS/Xray-core#166

### v2 UNSPEC

When this header is received:

```
0d 0a 0d 0a 00 0d 0a 51 55 49 54 0a 20 00 00 00
```

which is <v2 signature> + LOCAL command (\x20) + UNSPEC address family UNSPEC protocol (\x00)
`ErrUnsupportedAddressFamilyAndProtocol` is returned
https://github.com/pires/go-proxyproto/blob/adbbabed1f24a2123e1d4368ae8da8378b347643/v2.go#L85-L87
This is because UNSPEC is absent in supportedTransportProtocol 
https://github.com/pires/go-proxyproto/blob/adbbabed1f24a2123e1d4368ae8da8378b347643/addr_proto.go#L6-L23
This UNSPEC header is used when HAProxy doing health checks, or the proxied client is requesting via a unix socket.
Also the protocol spec says:

```
Only the UNSPEC protocol byte (\x00) is mandatory to implement on the receiver.
A receiver is not required to implement other ones, provided that it
automatically falls back to the UNSPEC mode for the valid combinations above
that it does not support.
```

### v1 PROXY UNKNOWN

When header `PROXY UNKNOWN\r\n` is received,
https://github.com/pires/go-proxyproto/blob/adbbabed1f24a2123e1d4368ae8da8378b347643/v1.go#L25-L33
returns ErrCantReadProtocolVersionAndCommand.
This is because in this case, there's only 2 parts separated by spaces, instead of 6 parts as `PROXY <PROTO> <SRC_ADDR> <DEST_ADDR> <SRC_PORT> <DEST_PORT>`
The protocol spec says:
```
If the announced transport protocol is "UNKNOWN", then the receiver knows that
the sender speaks the correct PROXY protocol with the appropriate version, and
SHOULD accept the connection and use the real connection's parameters as if
there were no PROXY protocol header on the wire.
```
Also, this kind of header can be formatted by the library.
https://github.com/pires/go-proxyproto/blob/adbbabed1f24a2123e1d4368ae8da8378b347643/v1.go#L76-L88

Although I really don't know much about Golang, I'll try my best to support you to resolve this issue. Thanks!

pires:
Fixes #60

I took the opportunity to improve test debuggability for v1 and v2 code.

Warning: changes to v1 errror handling may break someone's code but should be trivial to address. I'm taking this breaking change lightly because of that and because this is still a 0.x release, which should clearly communicate the API is not stable.

### DIFF
--- a/v2.go
+++ b/v2.go
@@ -82,8 +82,13 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		return nil, ErrCantReadAddressFamilyAndProtocol
 	}
 	header.TransportProtocol = AddressFamilyAndProtocol(b14)
-	if _, ok := supportedTransportProtocol[header.TransportProtocol]; !ok {
-		return nil, ErrUnsupportedAddressFamilyAndProtocol
+	//  UNSPEC is only supported when LOCAL is set.
+	if header.TransportProtocol == UNSPEC {
+		if header.Command != LOCAL {
+			return nil, ErrUnsupportedAddressFamilyAndProtocol
+		}
+		// Ignore everything else.
+		return header, nil
 	}
 
 	// Make sure there are bytes available as specified in length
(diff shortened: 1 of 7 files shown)


pires:
@emersion maybe you want to take a look?

bohanyang:
Sorry, seems that I found a bug... the server behind PP listener got the 15th and 16th `00 00` bytes, which is the length (for LOCAL UNSPEC it's zero)

v2.go
```diff
@@ -82,8 +82,13 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
...
 	}
 	header.TransportProtocol = AddressFamilyAndProtocol(b14)
-	if _, ok := supportedTransportProtocol[header.TransportProtocol]; !ok {
-		return nil, ErrUnsupportedAddressFamilyAndProtocol
+	//  UNSPEC is only supported when LOCAL is set.
+	if header.TransportProtocol == UNSPEC {
+		if header.Command != LOCAL {
+			return nil, ErrUnsupportedAddressFamilyAndProtocol
+		}
+		// Ignore everything else.
```
bohanyang, line 90: We should read 2 more bytes for the length

pires: If that's the case, then we should read all bytes if length > 0.

bohanyang: Yeah, I've opened a PR

PR #62, opened 2021-01-19
v2: read length and skip the bytes for UNSPEC

bohanyang:
Fixes https://github.com/pires/go-proxyproto/pull/61#issuecomment-762556537

### DIFF
--- a/v2.go
+++ b/v2.go
@@ -10,10 +10,15 @@ import (
 )
 
 var (
-	lengthV4   = uint16(12)
-	lengthV6   = uint16(36)
-	lengthUnix = uint16(216)
-
+	lengthUnspec      = uint16(0)
+	lengthV4          = uint16(12)
+	lengthV6          = uint16(36)
+	lengthUnix        = uint16(216)
+	lengthUnspecBytes = func() []byte {
+		a := make([]byte, 2)
+		binary.BigEndian.PutUint16(a, lengthUnspec)
+		return a
+	}()
 	lengthV4Bytes = func() []byte {
 		a := make([]byte, 2)
 		binary.BigEndian.PutUint16(a, lengthV4)
@@ -82,13 +87,9 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		return nil, ErrCantReadAddressFamilyAndProtocol
 	}
 	header.TransportProtocol = AddressFamilyAndProtocol(b14)
-	//  UNSPEC is only supported when LOCAL is set.
-	if header.TransportProtocol == UNSPEC {
-		if header.Command != LOCAL {
-			return nil, ErrUnsupportedAddressFamilyAndProtocol
-		}
-		// Ignore everything else.
-		return header, nil
+	// UNSPEC is only supported when LOCAL is set.
+	if header.TransportProtocol == UNSPEC && header.Command != LOCAL {
+		return nil, ErrUnsupportedAddressFamilyAndProtocol
 	}
 
 	// Make sure there are bytes available as specified in length
@@ -100,6 +101,12 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		return nil, ErrInvalidLength
 	}
 
+	// Return early if the length is zero, which means that
+	// there's no address information and TLVs present for UNSPEC.
+	if length == 0 {
+		return header, nil
+	}
+
 	if _, err := reader.Peek(int(length)); err != nil {
 		return nil, ErrInvalidLength
 	}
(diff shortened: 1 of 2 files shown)


v2.go
```diff
@@ -83,63 +88,67 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
...
-		}
-		header.SourceAddr = newIPAddr(header.TransportProtocol, addr.Src[:], addr.SrcPort)
-		header.DestinationAddr = newIPAddr(header.TransportProtocol, addr.Dst[:], addr.DstPort)
-	} else if header.TransportProtocol.IsUnix() {
-		var addr _addrUnix
-		if err := binary.Read(payloadReader, binary.BigEndian, &addr); err != nil {
-			return nil, ErrInvalidAddress
-		}
+	// Read addresses and ports for protocols other than UNSPEC
+	if header.TransportProtocol != UNSPEC {
```
pires, line 118: Please, add a comment stating if UNSPEC, this code, as the receiver, ignore address information and will skip straight to reading TLVs because length is greater than zero and there should be TLVs.

v2.go
```diff
@@ -83,63 +88,67 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
...
 		return nil, ErrCantReadLength
 	}
-	if !header.validateLength(length) {
+
+	// There's no need to validate for UNSPEC since length always >= 0
+	if header.TransportProtocol != UNSPEC && !header.validateLength(length) {
 		return nil, ErrInvalidLength
 	}
 
+	if length == 0 {
```
pires, line 106: Please, add a comment stating that if there are no TLVs present, length is zero and we can return early, otherwise, the code continues so TLVs reading is attempted.

v2.go
```diff
@@ -83,63 +88,67 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
...
 
 	// Make sure there are bytes available as specified in length
 	var length uint16
 	if err := binary.Read(io.LimitReader(reader, 2), binary.BigEndian, &length); err != nil {
 		return nil, ErrCantReadLength
 	}
-	if !header.validateLength(length) {
+
+	// There's no need to validate for UNSPEC since length always >= 0
+	if header.TransportProtocol != UNSPEC && !header.validateLength(length) {
```
pires, line 102: Let's keep header length validation in the func above, please, where we'll document that why UNSPEC is always >= 0 for future reference. At first, I didn't get it.

v2.go
```diff
@@ -159,7 +168,13 @@ func (header *Header) formatVersion2() ([]byte, error) {
 	buf.WriteByte(header.TransportProtocol.toByte())
 	// When UNSPEC, the receiver must ignore addresses and ports.
 	// Therefore there's no point in writing it.
```
pires, line 170: ...but if TLVs are present, those should be written. Please, augment this comment.

v2_test.go
```diff
@@ -136,6 +136,11 @@ var invalidParseV2Tests = []struct {
 		reader:        newBufioReader(append(append(append(SIGV2, byte(PROXY), byte(TCPv6)), lengthV6Bytes...), fixtureIPv4Address...)),
 		expectedError: ErrInvalidLength,
 	},
+	{
+		desc:          "UNSPEC but no length",
+		reader:        newBufioReader(append(SIGV2, byte(LOCAL), byte(UNSPEC))),
+		expectedError: ErrCantReadLength,
+	},
```
pires, line 143: Please, add a test with length greater than zero but no TLVs. This will improve code coverage.

v2_test.go
```diff
@@ -166,7 +171,7 @@ var validParseAndWriteV2Tests = []struct {
 	},
 	{
 		desc:   "local unspec",
-		reader: newBufioReader(append(append(SIGV2, byte(LOCAL), byte(TCPv4)), fixtureIPv4V2...)),
+		reader: newBufioReader(append(append(SIGV2, byte(LOCAL), byte(UNSPEC)), lengthUnspecBytes...)),
```
pires, line 174: Please, add a test with length greater than zero and matching TLVs.

pires (changes requested):
Nice catch!

v2.go
```diff
@@ -83,63 +88,67 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 	}
 	header.TransportProtocol = AddressFamilyAndProtocol(b14)
 	//  UNSPEC is only supported when LOCAL is set.
-	if header.TransportProtocol == UNSPEC {
-		if header.Command != LOCAL {
-			return nil, ErrUnsupportedAddressFamilyAndProtocol
-		}
-		// Ignore everything else.
-		return header, nil
+	if header.TransportProtocol.IsUnspec() && (header.Command != LOCAL || header.TransportProtocol != UNSPEC) {
```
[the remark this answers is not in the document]
bohanyang: This condition eusures that only AF_UNSPEC with transport protocol UNSPEC is accepted, and
for any other combinations of either AF_UNSPEC or TP UNSPEC, ErrUnsupportedAddressFamilyAndProtocol is returned.
for example
| AF (high 4 bit) | Transport Protocol (low 4 bit) | OK |
| --- | --- | --- |
| AF_UNSPEC | UNSPEC | OK |
| AF_INET | UNSPEC | ERR |
| AF_INET6 | UNSPEC | ERR |
| AF_UNSPEC | STREAM | ERR |

(should we do this?)

And also the command should be LOCAL.

v2.go
bohanyang: Ahh, sorry I forgot to change this comment

v2.go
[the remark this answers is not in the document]
pires: Maybe the following code needs to evolve to only return true if `ap == 0xFF`
https://github.com/pires/go-proxyproto/blob/22bc6148fd29e67df91deb259efc1980c29bc4e5/addr_proto.go#L41-L44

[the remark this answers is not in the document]
pires: As it is right now, it seems the `header.TransportProtocol.IsUnspec()` is redundant.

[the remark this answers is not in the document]
bohanyang: OK, I reverted this change for now

v2_test.go
```diff
@@ -221,6 +232,18 @@ var validParseAndWriteV2Tests = []struct {
 			rawTLVs:           fixtureTLV,
 		},
 	},
+	{
```
pires, line 235: Seems we're missing a test with UNSPEC, 0 length and no TLVs. That should address the code coverage decrease coveralls is complaining about.

pires (commented):
One last thing and I think we're good to go. Thank you very much for your patience and contribution!

v2_test.go
bohanyang: I think I've added (changed) it already. The test with `desc:   "local unspec"`, originally had header LOCAL TCPv4 fixtureIPv4V2, which seems like a copy-paste typo

bohanyang: Just added a test for UNSPEC to tlvFormatTests, and the Coveralls is now happy :)

v2_test.go
```diff
@@ -136,6 +142,16 @@ var invalidParseV2Tests = []struct {
...
+		desc:          "unspec length greater than zero but no TLVs",
+		reader:        newBufioReader(append(append(SIGV2, byte(LOCAL), byte(UNSPEC)), fixtureUnspecTLV[:2]...)),
+		expectedError: ErrInvalidLength,
+	},
+	{
+		desc:          "unspec length greater than zero but no TLVs",
+		reader:        newBufioReader(append(append(SIGV2, byte(LOCAL), byte(UNSPEC)), fixtureUnspecTLV[:2]...)),
+		expectedError: ErrInvalidLength,
+	},
```
pires, line 154: The test above is a duplicate or haven't I had enough caffeine yet? 😅

bohanyang: Sorry, removed!

bohanyang:
Just found that there're two tests with desc `TCPv4 length zero but with address and ports`, should I fix the desc or the fixture header?

pires:
Nice catch! Please, fix the desc of the last one which is clearly a test for TCPv6.

bohanyang:
Done!

pires (approved):
What a great contribution! Thank you very much, @bohanyang

pires:
https://github.com/pires/go-proxyproto/releases/tag/v0.4.1

PR #64, opened 2021-01-27
v1: fix command always LOCAL

bohanyang:
The changes in #61 overrides all `header.Command` to `LOCAL` (which was set to `PROXY` by `initVersion1`) unconditionally.

https://github.com/pires/go-proxyproto/blob/b6f440c01c68bf65792f463adcc71eb620fb0484/v1.go#L55-L60

This makes these functions returning the original addresses rather than the addresses contained in the proxy header.

https://github.com/pires/go-proxyproto/blob/b6f440c01c68bf65792f463adcc71eb620fb0484/protocol.go#L127-L155

Unfortunately, the problem was not revealed, since the `EqualsTo` function, which is used by the tests to compare actual and expected headers, straightly returns true when the command is LOCAL, and the command is not being compared.

https://github.com/pires/go-proxyproto/blob/b6f440c01c68bf65792f463adcc71eb620fb0484/header.go#L150-L170

This PR also updated the condition that checks the token length, made it to make sense.

The comments for ProtocolVersionAndCommand are also updated, made it less confusing.

### DIFF
--- a/version_cmd.go
+++ b/version_cmd.go
@@ -1,10 +1,16 @@
 package proxyproto
 
-// ProtocolVersionAndCommand represents proxy protocol version and command.
+// ProtocolVersionAndCommand represents the command in proxy protocol v2.
+// Command doesn't exist in v1 but it should be set since other parts of
+// this library may rely on it for determining connection details.
 type ProtocolVersionAndCommand byte
 
 const (
+	// LOCAL represents the LOCAL command in v2 or UNKNOWN transport in v1,
+	// in which case no address information is expected.
 	LOCAL ProtocolVersionAndCommand = '\x20'
+	// PROXY represents the PROXY command in v2 or transport is not UNKNOWN in v1,
+	// in which case valid local/remote address and port information is expected.
 	PROXY ProtocolVersionAndCommand = '\x21'
 )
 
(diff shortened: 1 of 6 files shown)


version_cmd.go
```diff
@@ -1,10 +1,17 @@
 package proxyproto
 
-// ProtocolVersionAndCommand represents proxy protocol version and command.
+// ProtocolVersionAndCommand represents the command in proxy protocol v2.
+// Command doesn't exist in v1 but it should be set
+// since other parts of this library may rely on it
+// for determining connection details.
```
pires, line 6: Can you please align the 3 lines above to the first one?

version_cmd.go
```diff
@@ -1,10 +1,17 @@
...
 type ProtocolVersionAndCommand byte
 
 const (
+	// LOCAL represents the LOCAL command in v2 or UNKNOWN protocol in v1,
```
pires, line 10:
```suggestion
	// LOCAL represents the LOCAL command in v2 or UNKNOWN transport in v1,
```

version_cmd.go
```diff
@@ -1,10 +1,17 @@
...
+	// in which case no address information is expected.
 	LOCAL ProtocolVersionAndCommand = '\x20'
+	// PROXY represents the PROXY command in v2 or protocols other than UNKNOWN in v1,
```
pires, line 13:
```suggestion
	// PROXY represents the PROXY command in v2 or transport is not UNKNOWN in v1,
```

version_cmd.go
```diff
@@ -13,17 +20,19 @@ var supportedCommand = map[ProtocolVersionAndCommand]bool{
 	PROXY: true,
 }
 
-// IsLocal returns true if the protocol version is \x2 and command is LOCAL, false otherwise.
+// IsLocal returns true if the command in v2 is LOCAL or the protocol in v1 is UNKNOWN,
```
pires, line 23:
```suggestion
// IsLocal returns true if the command in v2 is LOCAL or the transport in v1 is UNKNOWN,
```

version_cmd.go
```diff
@@ -13,17 +20,19 @@ var supportedCommand = map[ProtocolVersionAndCommand]bool{
...
+// i.e. when no address information is expected, false otherwise.
 func (pvc ProtocolVersionAndCommand) IsLocal() bool {
-	return 0x20 == pvc&0xF0 && 0x00 == pvc&0x0F
+	return LOCAL == pvc
 }
 
-// IsProxy returns true if the protocol version is \x2 and command is PROXY, false otherwise.
+// IsProxy returns true if the command in v2 is PROXY or the protocol in v1 is not UNKNOWN,
```
pires, line 29:
```suggestion
// IsProxy returns true if the command in v2 is PROXY or the transport in v1 is not UNKNOWN,
```

pires (commented):
Overall LGTM, but left a couple nits.

bohanyang:
Done fixing the comments!

pires (approved):
Thanks a lot for this contribution.

pires:
https://github.com/pires/go-proxyproto/releases/tag/v0.4.2

PR #67, opened 2021-02-16
Add PP2_TYPE_UNIQUE_ID

TimWolla:
This PR registers the unique ID TLV which was added to the PROXY spec on March,
5th 2020.


pires (approved):
Thanks a lot for your contribution, Tim!

PR #70, opened 2021-03-01
check v1 header bounds before reading

### LINKED ISSUE #69: parseVersion1() is not secure
```
func parseVersion1(reader *bufio.Reader) (*Header, error) {
	// Read until LF shows up, otherwise fail.
	// At this point, can't be sure CR precedes LF which will be validated next.
	line, err := reader.ReadString('\n')
```

The `reader` is a default `bufio.Reader` wrapping a `net.Conn`. It will read from the connection until it finds a newline. Since no limits are implemented in the code, a deliberately malformed V1 header could be used to exhaust memory in a server process using this code - a form of DDoS. The exploit is simple: send a stream starting with "PROXY" and keep sending data (which does not contain a newline) until the target stops acknowledging.

In most real world circumstances, the actual risk is small since only trusted sources should be allowed to send proxy protocol headers. However, this is still a security issue and should be resolved.

Easiest fix: `reader.Peek(107)` and scan for a newline. If none is found, then it is not a valid version 1 header anyway, so you can fail fast (the maximum v1 header size is 107 bytes). Otherwise, proceed with the `reader.ReadString('\n')` in full confidence.

isedev:
Resolves #69.

### DIFF
--- a/v1.go
+++ b/v1.go
@@ -3,6 +3,7 @@ package proxyproto
 import (
 	"bufio"
 	"bytes"
+	"io"
 	"net"
 	"strconv"
 	"strings"
@@ -24,7 +25,7 @@ func initVersion1() *Header {
 func parseVersion1(reader *bufio.Reader) (*Header, error) {
 	// The header must be terminated by CRLF and cannot be more than 107 bytes long.
 	buf, err := reader.Peek(107)
-	if err != nil {
+	if err != nil && err != io.EOF {
 		return nil, ErrLineMustEndWithCrlf
 	}
 	pos := bytes.Index(buf, []byte(crlf))
--- a/v1.go
+++ b/v1.go
@@ -22,17 +22,20 @@ func initVersion1() *Header {
 }
 
 func parseVersion1(reader *bufio.Reader) (*Header, error) {
-	// Read until LF shows up, otherwise fail.
-	// At this point, can't be sure CR precedes LF which will be validated next.
-	line, err := reader.ReadString('\n')
+	// The header must be terminated by CRLF and cannot be more than 107 bytes long.
+	buf, err := reader.Peek(107)
 	if err != nil {
 		return nil, ErrLineMustEndWithCrlf
 	}
-	if !strings.HasSuffix(line, crlf) {
+	pos := bytes.Index(buf, []byte(crlf))
+	if pos == -1 {
 		return nil, ErrLineMustEndWithCrlf
 	}
+	// Consume the input up to the terminator.
+	line := string(buf[:pos])
+	reader.Discard(pos + len(crlf))
 	// Check full signature.
-	tokens := strings.Split(line[:len(line)-2], separator)
+	tokens := strings.Split(line, separator)
 
 	// Expect at least 2 tokens: "PROXY" and the transport protocol.
 	if len(tokens) < 2 {


pires (commented):
Can you please add tests that prove the buffer growing unchecked as a first commit then this commit that fixes it?

v1.go
```diff
@@ -22,17 +23,20 @@ func initVersion1() *Header {
 }
 
 func parseVersion1(reader *bufio.Reader) (*Header, error) {
-	// Read until LF shows up, otherwise fail.
-	// At this point, can't be sure CR precedes LF which will be validated next.
-	line, err := reader.ReadString('\n')
-	if err != nil {
+	// The header must be terminated by CRLF and cannot be more than 107 bytes long.
+	buf, err := reader.Peek(107)
+	if err != nil && err != io.EOF {
```
pires, line 28: Per docs:
> If Peek returns fewer than n bytes, it also returns an error explaining why the read is short.
Is the error `io.EOF`? It's unclear from the docs and the code didn't immediately tell me it is.

v1.go
```diff
@@ -22,17 +23,20 @@ func initVersion1() *Header {
...
 		return nil, ErrLineMustEndWithCrlf
```
pires, line 29: Please, add a new error here, something like `ErrLineSizeOverflow=errors.New("proxyproto: header is invalid, must be of size <= 108 bytes")`

pires: This is to be captured in the tests I asked for.

v1.go
```diff
@@ -22,17 +23,20 @@ func initVersion1() *Header {
 }
 
 func parseVersion1(reader *bufio.Reader) (*Header, error) {
-	// Read until LF shows up, otherwise fail.
-	// At this point, can't be sure CR precedes LF which will be validated next.
-	line, err := reader.ReadString('\n')
-	if err != nil {
+	// The header must be terminated by CRLF and cannot be more than 107 bytes long.
```
pires, line 26: Please, add a note here justifying the _107 bytes long_, such as per spec
> (...) a 108-byte buffer is always enough to store all the line and a trailing zero for string processing.

v1.go
```diff
@@ -22,17 +23,20 @@ func initVersion1() *Header {
...
+	buf, err := reader.Peek(107)
+	if err != nil && err != io.EOF {
 		return nil, ErrLineMustEndWithCrlf
 	}
-	if !strings.HasSuffix(line, crlf) {
+	pos := bytes.Index(buf, []byte(crlf))
+	if pos == -1 {
```
pires, line 32: The existing tests already capture this.

pires (changes requested):
Please, add tests that prove the buffer growing unchecked as a first commit then this commit that fixes it?

v1.go
isedev: So there could be different errors - e.g. network read error. However, io.EOF is returned when there is less than 107 bytes available in the stream at that time, based on my tests. I will add some unit tests to cover various cases to validate this.

v1.go
isedev: ok, will do.

v1.go
isedev: ok

v1.go
isedev: Not clear what you mean by this. The `Peek` can return a buffer without CRLF at this point. It can also return a buffer that extends beyond the CRLF (e.g. short header, but more bytes available in the stream). So `HasSuffix` does not work here. Perhaps I've misunderstood your point though...

isedev:
I'll close this PR as the suggested updates don't resolve all problems and I want to base the fix on the commits in #71. I'll submit new fixes shortly.

v1.go
pires: Just a comment that existing tests (in main branch) already capture a scenario where the string read has LF but not CRLF.

PR #71, opened 2021-03-03
tests and fixes for issue #69

### LINKED ISSUE #69: parseVersion1() is not secure

isedev:
2As suggested in PR#70, a new pull request which captures just the tests proving issue #69, with some refinements:

- header overflow (i.e. reads more than the bufio.Reader default buffer size of 4096).
- "slow loris" type vulnerability, where the header read blocks waiting for terminator.

The second issue above can be used for DoS by consuming server process sockets.

Will then submit fixes - further testing shows that pull request #70 does not solve the problem properly.

Fixes #69

### DIFF
--- a/v1.go
+++ b/v1.go
@@ -3,6 +3,7 @@ package proxyproto
 import (
 	"bufio"
 	"bytes"
+	"fmt"
 	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


isedev:
Added the new fix. Let me know what you think.

v1.go
```diff
@@ -22,17 +22,79 @@ func initVersion1() *Header {
...
+	//
+	// ISSUE #69
+	// We can't use Peek here as it will block trying to fill the buffer, which
+	// will never happen if the header is TCP4 or TCP6 (max. 56 and 104 bytes
+	// respectively) and the server is expected to speak first.
+	//
+	// Similarly, we can't use ReadString or ReadBytes as these will keep reading
+	// until the delimiter is found; an abusive client could easily disrupt a
+	// server by sending a large amount of data that do not contain a LF byte.
+	// Another means of attack would be to start connectons and simply not send
```
pires, line 46: typo: _connectons_

v1.go
```diff
@@ -22,17 +22,79 @@ func initVersion1() *Header {
...
+	//    implementation constraints may not always easily permit this.
+	//
+	// We are subject to such implementation constraints. So we return an error if
+	// the header cannot be fully extracted with a single read of the underlying
+	// reader.
+	buf := make([]byte, 0, 107)
+	for {
+		b, err := reader.ReadByte()
+		if err != nil {
+			return nil, ErrCantReadVersion1Header
```
pires, line 72: I'd prefer for `err` to be wrapped otherwise we'll swallow it and the user will not understand what happened. I understand we aren't able to compare later in tests and I'm fine with it.

pires (changes requested):
The code mostly LGTM. Other than two comments I kindly request you tidy up the commit log.

I'd also like to compensate you for this contribution as it seems a serious security issue. Can you please reach out over Twitter (@therealpires) or email pjpires at googlemail?

Thanks a ton!

pires:
cc @emersion @mschneider82 @TimWolla @jefferai (I wish I remembered everyone's name/handle). Someone at Snyk is also willing to help getting a CVE for this which will be a first for me. Thanks everybody out there paying attention and willing to help.

TimWolla:
Thanks for the ping! I'm afraid I won't be much of a help here, though. I'm not even a go programmer. My contribution regarding the unique ID TLV was motivated by the fact that I was the person that added this TLV to the PROXYv2 spec (I'm somewhat involved with upstream HAProxy) and I just searched for proxy protocol libraries to make sure they are up to date.

v1.go
```diff
@@ -22,17 +22,79 @@ func initVersion1() *Header {
...
+	//    (..) The sender must always ensure that the header is sent at once, so that
+	//    the transport layer maintains atomicity along the path to the receiver. The
+	//    receiver may be tolerant to partial headers or may simply drop the connection
+	//    when receiving a partial header. Recommendation is to be tolerant, but
+	//    implementation constraints may not always easily permit this.
+	//
+	// We are subject to such implementation constraints. So we return an error if
+	// the header cannot be fully extracted with a single read of the underlying
+	// reader.
+	buf := make([]byte, 0, 107)
```
jefferai, line 68: I think you're better off using https://golang.org/pkg/io/#LimitReader rather than this loop.

jefferai: Never mind; I didn't see that the header isn't always 107 bytes.

v1.go
jefferai: You could compare in tests with `errors.Is`

v1.go
```diff
@@ -22,17 +22,79 @@ func initVersion1() *Header {
...
+		if reader.Buffered() == 0 {
+			// Header was not buffered in a single read. Since we can't
+			// differentiate between genuine slow writers and DoS agents,
+			// we abort. On healthy networks, this should never happen.
+			return nil, ErrCantReadVersion1Header
+		}
+	}
+
+	// Check for CR before LF.
+	if buf[len(buf)-2] != '\r' {
```
jefferai, line 92: This can panic if all that we have received is a newline as that would cause an index of `-2`. It needs a check to ensure that the length is at least 2 (it's clearly an error if not) before running this check.

jefferai:
@pires Thanks for pinging me! This generally looks good, left a couple of comments.

v1.go
isedev: That's a core assumption... `parseVersion1` is called all after the "PROXY" prefix is checked in `Read`.

isedev:
```
		if bytes.Equal(signature[:5], SIGV1) {
			return parseVersion1(reader)
		}
```

Take your point though, doesn't hurt to check and be safe.

isedev:
> The code mostly LGTM. Other than two comments I kindly request you tidy up the commit log.
Any suggestions how I tidy up the commit log? I agree needs to be done, just no idea how to without new pull request. If that's what it takes, more than happy to.

TimWolla:
> Any suggestions how I tidy up the commit log?

Use an interactive rebase `git rebase -i` and then force push to `isedev:feature/issue-69-tests`.

isedev:
> > Any suggestions how I tidy up the commit log?
> 
> Use an interactive rebase `git rebase -i` and then force push to `isedev:feature/issue-69-tests`.

Thanks. Learn something new everyday :)

v1.go
pires: Yeah but it won't work w/ wrapped errors, eg `ErrCantReadVersion1Header` wrapping an `io.EOF`, right? The idea is to tell the user why it failed and not swallow it.

pires (approved):
LGTM, thanks a tonne!

mschneider82:
@pires  can you tag a new release?

pires:
Working on it. Want to get Go 1.16 in and some fixes to tests I spotted while doing so.

mschneider82:
i do a pull req for traefik, where i used this lib to provide proxy protocol support, but traefik still depends on go 1.15, so requiring 1.16 here would break it. I will stick on last commit for now

v1.go
jefferai: Ah, thanks, I didn't see that. But yeah, why not add a check :-)

pires:
There's no such thing as enforcing Go 1.16 requirement. CI compiles and tests for Go 1.14 and Go 1.15, and now want to add 1.16, that's all.

mschneider82:
Ok fine. The traefik maintainers requested a new version tag

pires:
OK the tests fixing will have to wait as I need more time to find a reasonable strategy while also refactoring to reduce lots of duplicated code.

pires:
https://github.com/pires/go-proxyproto/releases/tag/v0.5.0

PR #74, opened 2021-04-10
Add support for ReadHeaderTimeout

unmarshal:
Set a read deadline when waiting for the PROXY protocol header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,38 @@ func main() {
 }
 ```
 
+### HTTP Server
+```go
+package main
+
+import (
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/pires/go-proxyproto"
+)
+
+func main() {
+	server := http.Server{
+		Addr: ":8080",
+	}
+
+	ln, err := net.Listen("tcp", server.Addr)
+	if err != nil {
+		panic(err)
+	}
+
+	proxyListener := &proxyproto.Listener{
+		Listener:          ln,
+		ReadHeaderTimeout: 10 * time.Second,
+	}
+	defer proxyListener.Close()
+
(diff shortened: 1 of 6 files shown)


pires:
I'm under the impression I'm about to learn something here but how can the coverage increase if no tests were implemented? Are the examples added here considered example tests?

unmarshal:
I was scratching my head about the test coverage too. I ended up just removing the examples and moving the https server to the README. I didn't want to restructure your project.

pires:
No, no, no, please, keep the code in the `examples` folder or let's make sure they are testable examples which is so much better than keeping examples in README.md. I am just curious how the hell goveralls reports a 0.7% increment in code coverage 😂

unmarshal:
Sounds good. Will do.

unmarshal:
code coverage didn't increase because of the examples. guessing it has to do with the new lines added when the deadline is set.

pires (changes requested):
Overall LGTM but please remove the binaries (maybe you added them accidentally) and add tests that show that an error is thrown when a client doesn't write the header after a defined amount of time.

unmarshal:
Whoops. Yes, that was a mistake. I will fix that and also add a test. Cheers.

unmarshal:
Sorry about the premature squashing and force pushes. Didn't realize it messed with github's review system.

pires (approved):
This is great, thank you very much!

jefferai:
@pires I strongly urge you to pick a suitable default for when no read timeout is set. Otherwise you put the onus on library users to know that they have to do this in order to be safe.

pires:
Agreed.

PR #78, opened 2021-07-27
Check Local Connection in parseVersion2

bobbybho:
If Connection is LOCAL then we should just skip the rest of the header as per spec. When a sender presents a LOCAL connection, it should not present any address so it sets this field to zero. Receivers MUST always consider this field to skip the appropriate number of bytes and must not assume zero is presented for LOCAL connections.

### DIFF
--- a/v2.go
+++ b/v2.go
@@ -111,6 +111,18 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		return nil, ErrInvalidLength
 	}
 
+	// if local then we should just skip the rest of the header as per spec
+	// 	When a sender presents a LOCAL connection, it should not present any
+	// address so it sets this field to zero. Receivers MUST always consider
+	// this field to skip the appropriate number of bytes and must not assume
+	// zero is presented for LOCAL connections.
+	if header.Command.IsLocal() {
+		if _, err := reader.Discard(int(length)); err != nil {
+			return nil, ErrInvalidAddress
+		}
+		return header, nil
+	}
+
 	// Length-limited reader for payload section
 	payloadReader := io.LimitReader(reader, int64(length)).(*io.LimitedReader)
 
@@ -230,7 +242,9 @@ func (header *Header) formatVersion2() ([]byte, error) {
 }
 
 func (header *Header) validateLength(length uint16) bool {
-	if header.TransportProtocol.IsIPv4() {
+	if header.Command.IsLocal() {
+		return true
+	} else if header.TransportProtocol.IsIPv4() {
 		return length >= lengthV4
 	} else if header.TransportProtocol.IsIPv6() {
 		return length >= lengthV6
--- a/v2.go
+++ b/v2.go
@@ -111,9 +111,8 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		return nil, ErrInvalidLength
 	}
 
-	// if local then we should just skip the rest of the header as per spec
-	// 	When a sender presents a LOCAL connection, it should not present any
-	// address so it sets this field to zero. Receivers MUST always consider
+	// When a sender presents a LOCAL connection, it should not present any address
+	// so it sets this field to zero. Receivers MUST always consider
 	// this field to skip the appropriate number of bytes and must not assume
 	// zero is presented for LOCAL connections.
 	if header.Command.IsLocal() {
--- a/v2.go
+++ b/v2.go
@@ -120,6 +120,7 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		if _, err := reader.Discard(int(length)); err != nil {
 			return nil, ErrInvalidAddress
 		}
+
 		return header, nil
 	}
 
@@ -180,6 +181,12 @@ func (header *Header) formatVersion2() ([]byte, error) {
 	buf.Write(SIGV2)
 	buf.WriteByte(header.Command.toByte())
 	buf.WriteByte(header.TransportProtocol.toByte())
+
+	if header.Command.IsLocal() {
+		buf.Write([]byte{'\x00', '\x00'})
+		return buf.Bytes(), nil
+	}
+
 	if header.TransportProtocol.IsUnspec() {
 		// For UNSPEC, write no addresses and ports but only TLVs if they are present
 		hdrLen, err := addTLVLen(lengthUnspecBytes, len(header.rawTLVs))
--- a/v2_test.go
+++ b/v2_test.go
@@ -171,13 +171,13 @@ var validParseAndWriteV2Tests = []struct {
 			Version:           2,
 			Command:           LOCAL,
 			TransportProtocol: TCPv4,
-			SourceAddr:        v4addr,
-			DestinationAddr:   v4addr,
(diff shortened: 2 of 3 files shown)


igor-kupczynski:
Hey @pires (sorry for the direct ping) any chance this may get reviewed and approved? I understand this is an OSS project and we have no expectation of anything, just want to understand where do we stand with this. We keep our internal private fork and wanted to upstream some fixes and maybe switch back to the upstream.

v2.go
```diff
@@ -111,6 +111,19 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		return nil, ErrInvalidLength
 	}
 
+	// if local then we should just skip the rest of the header as per spec
+	// 	When a sender presents a LOCAL connection, it should not present any
```
pires, line 115: Please, fix the formatting.

v2.go
```diff
@@ -168,6 +181,12 @@ func (header *Header) formatVersion2() ([]byte, error) {
 	buf.Write(SIGV2)
 	buf.WriteByte(header.Command.toByte())
 	buf.WriteByte(header.TransportProtocol.toByte())
+
```
pires, line 184:
```suggestion
// When a sender presents a LOCAL connection, it should not present any address
// so it sets this field (source and destination addresses) to zero.
```

v2.go
```diff
@@ -111,6 +111,19 @@ func parseVersion2(reader *bufio.Reader) (header *Header, err error) {
 		return nil, ErrInvalidLength
 	}
 
+	// if local then we should just skip the rest of the header as per spec
+	// 	When a sender presents a LOCAL connection, it should not present any
+	// address so it sets this field to zero. Receivers MUST always consider
+	// this field to skip the appropriate number of bytes and must not assume
+	// zero is presented for LOCAL connections.
```
pires, line 118: In the comment above, please keep only the relevant pieces, namely what happens for the receiver side.

v2_test.go
```diff
@@ -483,17 +484,6 @@ var tlvFormatTests = []struct {
 			rawTLVs:           make([]byte, 1<<16),
 		},
 	},
-	{
-		desc: "local unspec",
```
pires, line 487: Why is this test deleted? IIUC only addresses are ignored and attributes like TLVs and Transport Protocol.

v2_test.go
```diff
@@ -241,7 +243,6 @@ var validParseAndWriteV2Tests = []struct {
 			TransportProtocol: UNSPEC,
 			SourceAddr:        nil,
 			DestinationAddr:   nil,
-			rawTLVs:           fixtureTLV,
```
pires, line 244: Same here: IIUC only addresses are ignored and attributes like TLVs and Transport Protocol.

pires:
Sorry, @igor-kupczynski, for some reason this slipped. Thanks for your contributions and gentle ping!

v2_test.go
bobbybho: @pires The reason I delete this test case is that in our unit test cases (based on packets received from AWSLB), if local connection is presented, the rest of the headers and TLVs are ignored. That is why in v2.go, we added a check to see if the header is local, and if so, the function just returns the header without coping the remaining headers and the TLV data. 

I reviewed the latest proxy protocol spec https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt but couldn't find any information indicating whether the receiver should or should not keep the TLV data if the connection is local. We assume that the remaining headers and TLVs are ignored in this case. cc @igor-kupczynski to get some feedback.

pires: The spec is clear that when LOCAL is presented, the receiver is to ignore the address information, nothing else. Implicitly, you can still pass LOCAL+<INET FAMILY>+TLVs so let's keep that until the standard makes it clear. I also don't see any problem in keeping this as this information is only processed if it's there.

bobbybho: @pires Thank you for your feedback. Just want to provide a little bit more context. The reason we ignore the TLVs if it is local connection because we captured the following packets from AWS and Azure:

AWS:
```
awsLocalPP := []byte{
		'\x0d', '\x0a', '\x0d', '\x0a', '\x00', '\x0d', '\x0a', '\x51', '\x55', '\x49', '\x54', '\x0a', //v2 sig
		'\x20',         // command LOCAL
		'\x11',         // Address Family TCP over IPv4
		'\x00', '\x54', // length 84 bytes
		'\x03', '\x00', '\x04', '\xbc', '\xe5', '\x95', '\x1f', // This is actually a CRC type 3 TLV
		'\x04', '\x00', '\x4a', // This is a NOOP TLV which covers the rest of the nul bytes in the header
		'\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00',
		'\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00',
		'\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00',
		'\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00',
		'\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00', '\x00',
		'\x00', '\x00', '\x00', '\x00',
	}
```
Azure:
```
azureLocalHealthPP := []byte{
		'\x0d', '\x0a', '\x0d', '\x0a', '\x00', '\x0d', '\x0a', '\x51', '\x55', '\x49', '\x54', '\x0a', // v2 sig
		'\x20',         // command LOCAL
		'\x00',         // address family ??
		'\x00', '\x0c', // length
		'\x00', '\x00', '\x00', '\x00',
		'\x00', '\x00', '\x00', '\x00',
		'\x00', '\x00',
		'\x00', '\x00',
	}
```
It seems that AWS sets the TLVs to all zeros if it is local connection and Azure drops the TLVs. That is why in our implementation, we ignore the TLV data. 

I've talked to @igor-kupczynski and Franek (another contributor), we also think that it should be ok to pass TLVs if the connection is local. I am going to run more tests to confirm.

I will provide an update once I have more test results and if everything is ok, we could probably close this PR.

pires: I'm waiting on the tests to be added again, just to be sure we're on the same page.

bobbybho: @pires Thank you for following with this PR. Currently I am running e2e integration tests on aws, gcp and azure. I don't have the test results yet and will provide an update once my testing is completed.

bobbybho:
@pires We have finished running integration tests in AWS, GCP and Azure and no error is found in the test result so far. As a result, I am going to close this PR since the latest upstream implementation is working with those cloud providers and we don't need to change the local connection behaviours. Thank you again for reviewing our PR and sharing your feedback with us.

pires:
Terrific! Thank you very much for your contributions and keeping me posted.

Issue #73, opened 2021-03-05
Tests goroutines call t.Fatalf


pires:
And `go vet` doesn't like it

```
Error: ./protocol_test.go:29:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:37:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:40:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:88:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:99:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:102:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:165:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:176:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:179:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:231:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:259:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:291:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:335:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:359:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:362:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:433:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:466:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:499:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:542:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:641:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:710:4: call to (*T).Fatalf from a non-test goroutine
Error: ./protocol_test.go:817:5: call to (*B).Fatalf from a non-test goroutine
Error: ./protocol_test.go:837:5: call to (*B).Fatalf from a non-test goroutine
Error: ./protocol_test.go:842:6: call to (*B).Fatalf from a non-test goroutine
[...]

PR #80, opened 2021-08-25
Update the way read header is handled

### LINKED ISSUE #75: ReadHeaderTimeout is setting a hard timeout on connections regardless of header sending
I think this setting is supposed to timeout connections that are not sending the headers and terminate them after this amount of time. However, I set the setting to 5 seconds, and the result is that all connections are terminated after 5 seconds, even if they correctly sent the headers. The connection starts up, but is then abruptly ended after the 5 seconds are passed.

I think what is missing in #74 is a call to `SetReadDeadline` to reset the timeout after the proxy header was sent successfully.

antoniomika:
edited by @pires 

Fixes #75

### DIFF
--- a/addr_proto.go
+++ b/addr_proto.go
@@ -15,32 +15,32 @@ const (
 
 // IsIPv4 returns true if the address family is IPv4 (AF_INET4), false otherwise.
 func (ap AddressFamilyAndProtocol) IsIPv4() bool {
-	return 0x10 == ap&0xF0
+	return ap&0xF0 == 0x10
 }
 
 // IsIPv6 returns true if the address family is IPv6 (AF_INET6), false otherwise.
 func (ap AddressFamilyAndProtocol) IsIPv6() bool {
-	return 0x20 == ap&0xF0
+	return ap&0xF0 == 0x20
 }
 
 // IsUnix returns true if the address family is UNIX (AF_UNIX), false otherwise.
 func (ap AddressFamilyAndProtocol) IsUnix() bool {
-	return 0x30 == ap&0xF0
+	return ap&0xF0 == 0x30
 }
 
 // IsStream returns true if the transport protocol is TCP or STREAM (SOCK_STREAM), false otherwise.
 func (ap AddressFamilyAndProtocol) IsStream() bool {
-	return 0x01 == ap&0x0F
+	return ap&0x0F == 0x01
 }
 
 // IsDatagram returns true if the transport protocol is UDP or DGRAM (SOCK_DGRAM), false otherwise.
 func (ap AddressFamilyAndProtocol) IsDatagram() bool {
-	return 0x02 == ap&0x0F
+	return ap&0x0F == 0x02
 }
 
 // IsUnspec returns true if the transport protocol or address family is unspecified, false otherwise.
 func (ap AddressFamilyAndProtocol) IsUnspec() bool {
-	return (0x00 == ap&0xF0) || (0x00 == ap&0x0F)
+	return (ap&0xF0 == 0x00) || (ap&0x0F == 0x00)
 }
 
 func (ap AddressFamilyAndProtocol) toByte() byte {
--- a/protocol.go
+++ b/protocol.go
@@ -8,10 +8,15 @@ import (
 	"time"
 )
 
+var DEFAULT_TIMEOUT = 200 * time.Millisecond
+
 // Listener is used to wrap an underlying listener,
 // whose connections may be using the HAProxy Proxy Protocol.
 // If the connection is using the protocol, the RemoteAddr() will return
-// the correct client address.
+// the correct client address. ReadHeaderTimeout will be applied to all
+// connections in order to prevent blocking operations. If no ReadHeaderTimeout
+// is set, a default of 200ms will be used. This can be disabled by setting the
+// timeout to < 0.
 type Listener struct {
 	Listener          net.Listener
 	Policy            PolicyFunc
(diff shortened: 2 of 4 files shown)


drakkan:
Hi,

I just tested this patch with [SFTPGo](https://github.com/drakkan/sftpgo)  for proxy SSH and FTP connections, it looks fine. Are there any blockers to merge it? thanks

protocol.go
```diff
@@ -30,6 +30,9 @@ type Conn struct {
 	ProxyHeaderPolicy Policy
 	Validate          Validator
 	readErr           error
+	ReadHeaderTimeout time.Duration
+	ReadDeadline      time.Time
+	WriteDeadline     time.Time
```
pires, line 35: Do we really need to export these last two attributes?

pires (commented):
Can you please add comments to the code explaining what you're doing here? It took me a bit to understand why we'd care about keeping track of deadline times for read and write, and I'm not even sure I got it right. IIUC we must keep track so that ReadHeaderTimeout doesn't interfere w/ user provided deadlines.

protocol.go
antoniomika: Nope! I'll update and add some documentation. Just wanted to add this as some boilerplate to show you what I was thinking but I'll get it merge ready.

antoniomika:
Hey @pires,

Added some more documentation, cleaned up some unneeded boilerplate and we should be ready to go!

Best,

protocol.go
```diff
@@ -11,7 +11,10 @@ import (
 // Listener is used to wrap an underlying listener,
 // whose connections may be using the HAProxy Proxy Protocol.
 // If the connection is using the protocol, the RemoteAddr() will return
-// the correct client address.
+// the correct client address. ReadHeaderTimeout will be applied to all
+// connections in order to prevent blocking operations. If no ReadHeaderTimeout
+// is set, a default of 200ms will be used. This can be disabled by setting the
+// timeout to < 0.
```
pires, line 17: I think we're missing a test for `ReadHeaderTimeout < 0`, right?

protocol.go
```diff
@@ -74,6 +74,15 @@ func (p *Listener) Accept() (net.Conn, error) {
 		WithPolicy(proxyHeaderPolicy),
 		ValidateHeader(p.ValidateHeader),
 	)
+
+	// If the ReadHeaderTimeout for the listener is 0, set a default of 200ms
+	if p.ReadHeaderTimeout == 0 {
+		p.ReadHeaderTimeout = 200 * time.Millisecond
```
pires, line 80: I think it could be helpful to export the default as a var so consumers of this library can override it. WDYT?

pires (commented):
This LGTM but I have two questions. Please, be patient w/ me here @antoniomika.

Thank you very, very much for this contribution!

protocol.go
antoniomika: Not sure what the benefit of that is over just setting the timeout itself. Do you have a use case that should work in?

protocol.go
antoniomika: Technically not as viewed from a coverage standpoint, but I can add a test for it. The reason I say not is because we'd expect the default to occur without setting deadlines, which is the read would hang forever.

protocol.go
antoniomika: Changed just for simplicity

protocol.go
antoniomika: Added a test and cleaned up some lints for golangci-lint

antoniomika:
@pires just a heads up, this should merge first, followed by #76 after this lands.

pires:
> @pires just a heads up, this should merge first, followed by #76 after this lands.

What is there in #76 that isn't solved here?

protocol.go
pires: For instance, one can override the default during compile time, eg `go build -ldflags -X "github.com/pires/go-proxyproto.DefaultReadHeaderTimeout=1s"` which would then apply too all uses of the lib, regardless of listener/conn.

addr_proto.go
```diff
@@ -15,32 +15,32 @@ const (
...
 // IsDatagram returns true if the transport protocol is UDP or DGRAM (SOCK_DGRAM), false otherwise.
 func (ap AddressFamilyAndProtocol) IsDatagram() bool {
-	return 0x02 == ap&0x0F
+	return ap&0x0F == 0x02
 }
 
 // IsUnspec returns true if the transport protocol or address family is unspecified, false otherwise.
 func (ap AddressFamilyAndProtocol) IsUnspec() bool {
-	return (0x00 == ap&0xF0) || (0x00 == ap&0x0F)
+	return (ap&0xF0 == 0x00) || (ap&0x0F == 0x00)
```
pires, line 43: Why was #82 merged into this? Please, remove.

drakkan:
> > @pires just a heads up, this should merge first, followed by #76 after this lands.
> 
> What is there in #76 that isn't solved here?

To apply this patch you have to apply #76 first.

protocol.go
drakkan: > Added a test and cleaned up some lints for golangci-lint

you included the same changes in #82, I'll close my PR

pires (commented):
Somehow #82 landed here and I'd like to give the chance for @drakkan to get their contribution in. Can you please remove this?

pires:
> > > @pires just a heads up, this should merge first, followed by #76 after this lands.
> > 
> > 
> > What is there in #76 that isn't solved here?
> 
> To apply this patch you have to apply #76 first.

Damn, I must have that espresso urgently! 😅

antoniomika:
> Somehow #82 landed here and I'd like to give the chance for @drakkan to get their contribution in. Can you please remove this?

Sorry about that, I should've checked if there was another PR with the static check fixes. I use code spaces and was a bit OCD with the linting issues. I actually started working on #73 but those changes were more complicated so I decided to leave them for when I had more time. You can remove those changes in your PR now and merge in #82.

pires:
Yeah #73 is painful. Also have a branch for a long while now 😅

PR #81, opened 2021-09-06
Add support for GCP extensions

igor-kupczynski:
```
The support is modeled on azure.go under tlvparse. The GCP spec is similar but
encodes uint64. There's no subtype. Since zero value of uint64 is technically
a valid header value, I've opted for an extra boolean to indicate if it was
found (same as with Azure).
```

Given that we already have AWS, and Azure in `tlvparse` I thought it will be a good idea to add the GCP support as well.

We use it with GCP in our pre-prod env and all seems to be working well.

Let me know if you think it makes sense to have this with the upstream lib.

### DIFF
--- a/tlvparse/gcp.go
+++ b/tlvparse/gcp.go
@@ -1,0 +1,47 @@
+package tlvparse
+
+import (
+	"encoding/binary"
+
+	"github.com/pires/go-proxyproto"
+)
+
+const (
+	// PP2_TYPE_GCP indicates a Google Cloud Platform header
+	PP2_TYPE_GCP proxyproto.PP2Type = 0xE0
+)
+
+// ExtractPSCConnectionID returns the first PSC Connection ID in the TLV if it exists and is well-formed and
+// a bool indicating one was found.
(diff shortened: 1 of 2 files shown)


igor-kupczynski:
I've noticed a naming inconsistency and force-pushed a fixed version. Hope that OK -- my thinking was that a simple rename doesn't add a lot to the history.

pires (approved):
The only thing I'd like to see here and maybe it's something for the future is to provide some structure, such as func names tied to PP2 type, eg `PP2GCP.ExtractPSCConnectionID`, or maybe have `tlvparse/{aws,azure,gcp}` packages so folks don't need to read the code to know which functions are available to them for a particular type.

That said, thank you very much for your contributions! 🙇🏻

igor-kupczynski:
Nice, thanks for merging :)  and thanks for maintaining the library.

> func names tied to PP2 type, eg `PP2GCP.ExtractPSCConnectionID`, or maybe have `tlvparse/{aws,azure,gcp}`

Agree, it would be nice to improve the consistency here. I'll think about it for a bit.

pires:
Released in https://github.com/pires/go-proxyproto/releases/tag/v0.6.1

Issue #83, opened 2021-09-15
DefaultReadHeaderTimeout is too strict


emersion:
200ms is too strict and is causing breakage downstream, especially when TLS is involved. Having a timeout is good, but should not penalize slow connections.

The stdlib is using a 10s timeout for the TLS handshake for instance.

Ref https://github.com/emersion/soju/pull/33

pires:
We didn't have any data to justify more or less and we made it possible to replace it. Actually, one of my requests was to export the default so it could be overridden at compile time too, eg `go build -ldflags -X "github.com/pires/go-proxyproto.DefaultReadHeaderTimeout=10s"`, so you don't need to set it per listener.

I'm fine increasing the default for less experienced consumers of the API to not run into problems early on.

csmith:
> eg `go build -ldflags -X "github.com/pires/go-proxyproto.DefaultReadHeaderTimeout=10s"`

As an aside, this doesn't work. -X only works with string vars:

```
 -X importpath.name=value
	Set the value of the string variable in importpath named name to value.
	This is only effective if the variable is declared in the source code either uninitialized
	or initialized to a constant string expression.
```
-- https://pkg.go.dev/cmd/link

pires:
TIL thanks a lot @csmith.

PR #84, opened 2021-09-15
Bump DefaultReadHeaderTimeout to 10s

emersion:
When dealing with slow networks and TLS, 200ms is often hit. Bump
the default timeout to 10s, which is what crupto/tls uses for the
TLS handshake.

Closes: https://github.com/pires/go-proxyproto/issues/83

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -12,7 +12,7 @@ import (
 // be read from the wire, if Listener.ReaderHeaderTimeout is not set.
 // It's kept as a global variable so to make it easier to find and override,
 // e.g. go build -ldflags -X "github.com/pires/go-proxyproto.DefaultReadHeaderTimeout=1s"
-var DefaultReadHeaderTimeout = 200 * time.Millisecond
+var DefaultReadHeaderTimeout = 10 * time.Second
 
 // Listener is used to wrap an underlying listener,
 // whose connections may be using the HAProxy Proxy Protocol.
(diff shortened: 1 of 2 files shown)


protocol.go
```diff
@@ -12,7 +12,7 @@ import (
 // be read from the wire, if Listener.ReaderHeaderTimeout is not set.
 // It's kept as a global variable so to make it easier to find and override,
 // e.g. go build -ldflags -X "github.com/pires/go-proxyproto.DefaultReadHeaderTimeout=1s"
-var DefaultReadHeaderTimeout = 200 * time.Millisecond
+var DefaultReadHeaderTimeout = 10 * time.Second
```
pires, line 15: Can you please add the reason here why this timeout? Or maybe just point to whatever const/var you mentioned is exposed by stdlib.

protocol_test.go
```diff
@@ -265,6 +265,8 @@ func TestReadHeaderTimeoutIsReset(t *testing.T) {
 // we expect the actual address and port to be returned,
 // rather than the ProxyHeader we defined.
 func TestReadHeaderTimeoutIsEmpty(t *testing.T) {
+	DefaultReadHeaderTimeout = 200 * time.Millisecond
```
pires, line 268: I think this may have impact on subsequent tests, so maybe override the listener timeout instead.

emersion: The point of the test is to check the default.

protocol.go
emersion: There's no other reason than "it's a reasonable timeout".

emersion:
Gentle ping

pires:
Sorry @emersion in between work and sickness, I just haven't been able to look into this. Thanks a lot for your contribution, once more!

pires (approved):
LGTM

Issue #86, opened 2022-02-11
Data race in SetDeadline


jjiang-stripe:
Hello!

I have a test that runs with the `-race` flag that started to fail with v0.6.1. I've created a minimal repro here: https://github.com/jjiang-stripe/proxyproto-race/blob/master/main.go

That example is just running a super basic SSH server and client and it looks like Go's SSH server's running into a race when trying to update deadlines in its `readLoop` and `kexLoop` goroutines:
```
$ go run -race main.go 
==================
WARNING: DATA RACE
Write at 0x00c0000da058 by goroutine 27:
  github.com/pires/go-proxyproto.(*Conn).SetDeadline()
      /Users/jjiang/go/pkg/mod/github.com/pires/go-proxyproto@v0.6.1/protocol.go:218 +0x4f
  github.com/gliderlabs/ssh.(*serverConn).updateDeadline()
      /Users/jjiang/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.3/conn.go:53 +0x2f9
  github.com/gliderlabs/ssh.(*serverConn).Write()
      /Users/jjiang/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.3/conn.go:18 +0x44
  bufio.(*Writer).Flush()
      /usr/local/Cellar/go/1.17.2/libexec/src/bufio/bufio.go:607 +0xf0
  golang.org/x/crypto/ssh.(*connectionState).writePacket()
      /Users/jjiang/go/pkg/mod/golang.org/x/crypto@v0.0.0-20210616213533-5ff15b29337e/ssh/transport.go:182 +0x12b
  golang.org/x/crypto/ssh.(*transport).writePacket()
      /Users/jjiang/go/pkg/mod/golang.org/x/crypto@v0.0.0-20210616213533-5ff15b29337e/ssh/transport.go:172 +0x98
  golang.org/x/crypto/ssh.(*handshakeTransport).pushPacket()
      /Users/jjiang/go/pkg/mod/golang.org/x/crypto@v0.0.0-20210616213533-5ff15b29337e/ssh/handshake.go:223 +0x84d
  golang.org/x/crypto/ssh.(*handshakeTransport).sendKexInit()
      /Users/jjiang/go/pkg/mod/golang.org/x/crypto@v0.0.0-20210616213533-5ff15b29337e/ssh/handshake.go:472 +0x84e
  golang.org/x/crypto/ssh.(*handshakeTransport).kexLoop()
[...]

PR #87, opened 2022-03-01
protocol: fixed deadline datarace #86

mschneider82:
Fixed the race by using atomic.Value instead of unprotected time.Time 

It was nessasarry to restructure the Conn struct because of this:

```struct of size 80 could be 72 (fieldalignment)```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
       run: go vet
 
     - name: Test
-      run: go test -v -covermode=count -coverprofile=coverage.out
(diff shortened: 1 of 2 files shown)


pires:
Thanks @mschneider82! Could you please _turn on_ race mode for testing to prove this works (as it looks it does)?

pires:
TIL https://medium.com/@sebassegros/golang-dealing-with-maligned-structs-9b77bacf4b97

mschneider82:
I have added a nil check for atomic.Value with default time.Time{} (no deadline timeout) because TestPassthrough failed. In case no deadline was set

mschneider82:
done, ready for review & merge

PR #91, opened 2022-04-28
Add method to get SSL cipher from the SSL TLV

dgl:
Tests are using a header captured from haproxy with "proxy-v2-options
ssl-cipher" set.

To use this with haproxy you also need to set the ssl-cipher option,
something like:

    backend test
      default-server send-proxy-v2-ssl proxy-v2-options ssl-cipher
      server a [::1]:8080


pires (approved):
LGTM thank you very much.

PR #92, opened 2022-04-29
Allow IPv6 compat addresses (4in6) when parsing TCPv6 in V1 header 

Freeaqingme:
This is a follow-up on MR #49 by @emersion . All credits should go to him :)

I had a need to parse 4in6 addresses myself as well, so I took his MR as a starting point and took it from there. I have not yet tested this code in production, but I figured I'd already turn it into a PR.

There is one issue that I had to sort of work around: In the default net.IP library, an IPv4 address is (or can be) stored more or less in the same way an IPv6 address is stored with a ::FFFF: prefix. As such, there's no way to make a distinction based on a net.IP object if it was an IPv4 address, or IPv6 4in6 address.

### DIFF
--- a/header_test.go
+++ b/header_test.go
@@ -13,20 +13,23 @@ import (
 // Stuff to be used in both versions tests.
 
 const (
-	NO_PROTOCOL   = "There is no spoon"
-	IP4_ADDR      = "127.0.0.1"
-	IP6_ADDR      = "::1"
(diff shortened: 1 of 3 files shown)


pires:
Maybe it's a good time to introduce https://pkg.go.dev/net/netip#Addr.Is4In6?

PR #96, opened 2023-03-01
Add SKIP policy to not expect a PROXY header

### LINKED ISSUE #95: CIDR based policy for a simple non-PROXY traffic
I believe I'm not the first one who tries to solve the PROXY PROTOCOL vs no PROXY PROTOCOL traffic problem for a single listener. This is a very common behavior for k8s clusters that use PROXY_PROTOCOL based loadbalancers.

In short, the k8s pods clients local traffic doesn't use PROXY PROTOCOL, therefore accessing the service locally that expects PROXY PROTOCOL fails. There are two dirty solutions for this: route local traffic to the external loadbalancer IP (to add the required header), or fallback due to timeout.

Both options above have flaws: LB adds latency and load, timeout adds latency and acts unpredictably on heavy load.

I know that it's possible to create an extra wrapper around the original Conn, but this would look dirty. I propose to add a new SKIP policy, which will treat a connection as a regular one and it won't trigger proxyproto handlers. The SKIP policy can be defined in a `PolicyFunc` callback.

Objections?

kayrus:
Resolves #95

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -32,7 +32,30 @@ const (
 	// a PROXY header is not present, subsequent reads do not. It is the task
 	// of the code using the connection to handle that case properly.
 	REQUIRE
+	// SKIP accepts a connection without requiring the PROXY header
+	// Note: an example usage can be found in the SkipProxyHeaderForCIDR
+	// function.
+	SKIP
 )
+
+// SkipProxyHeaderForCIDR returns a PolicyFunc which can be used to accept a
+// connection from a skipHeaderCIDR without requiring a PROXY header, e.g.
+// Kubernetes pods local traffic. The def is a policy to use when an upstream
+// address doesn't match the skipHeaderCIDR.
+func SkipProxyHeaderForCIDR(skipHeaderCIDR *net.IPNet, def Policy) PolicyFunc {
+	return func(upstream net.Addr) (Policy, error) {
+		ip, err := ipFromAddr(upstream)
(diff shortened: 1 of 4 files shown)


kayrus:
@pires could you please take a look on this PR?

pires:
Please, add tests and documentation. I think what you wrote in #95 description is a good starting point.

kayrus:
@pires where should I put the documentation?

kayrus:
@pires is a `SKIP accepts a connection without approaching a PROXY header` comment looks better than `SKIP accepts a connection without a PROXY header`

kayrus:
@pires is there anything else I need to add to the PR?

pires:
I need a couple days to review this. Thank you very much for your contribution.

pires:
Sorry @kayrus but with a recent bump in the Go tooling version(s), tests that spawn go-routines fail to pass `go vet` if they have calls to `Fatalf` and such. Can you, please, work on that?

Other than that, this LGTM!

pires (changes requested):
Sorry @kayrus but with a recent bump in the Go tooling version(s), tests that spawn go-routines fail to pass `go vet` if they have calls to `Fatalf` and such. Can you, please, work on that?

Other than that, this LGTM!

kayrus:
@pires fixed

pires (approved):
Thanks a lot for your contribution.

PR #104, opened 2023-10-29
header.go prevent otherHeader.rawTLVs being nil

hilbix:
If header.Version != otherHeader.Version then otherHeader.rawTLVs is not present.

This resulted in a crash at my side.  Switching position of the 2 ifs solves it.


pires:
Thank you for contributing.

Can you, please, add a test that covers for your scenario, ie that triggers the panic?

pires (approved):

PR #108, opened 2024-04-18
Add support for validating the downstream ip of the connection

### LINKED ISSUE #107: [Feature] Support for policy on local address
Currently we can apply policy to verify the upstream address https://github.com/pires/go-proxyproto/blob/e5b291b295b4a6e0425a134c8b3bfa6b2e692866/protocol.go#L70-L81. There could be scenarios where there are multiple interfaces the server is listening on and we may want to enable the proxy protocol on only 1 interface. It would help to have policy support verifying the local address. 
I can create a PR if the feature is okay to be implemented.

kmala:
fixes #107


dims:
@pires can you please approve the CI workflow so we can check if the new test being added works fine? thanks!

pires:
@dims for you the world, thanks for pinging.

dims:
@kmala please review the coverage tests, looks like we need to make sure they don't regress.

kmala:
> please review the coverage tests, looks like we need to make sure they don't regress.

Update and verified locally that the coverage has increased.

policy.go
```diff
@@ -43,7 +44,7 @@ const (
 // Kubernetes pods local traffic. The def is a policy to use when an upstream
 // address doesn't match the skipHeaderCIDR.
 func SkipProxyHeaderForCIDR(skipHeaderCIDR *net.IPNet, def Policy) PolicyFunc {
-	return func(upstream net.Addr) (Policy, error) {
+	return func(upstream net.Addr, downstream net.Addr) (Policy, error) {
```
pires, line 47: This breaks the API and I'm thinking there's no need to do this. `upstream` is just a placeholder for any `net.Addr` so unless both upstream and downstream addresses have to be evaluated in a policy, I think we can just keep the API as is, maybe rename `upstream` to `address` and the new `PolicyFunc` just works. WDYT?

kmala: I thought about it and the issue i faced is where this function is called https://github.com/pires/go-proxyproto/blob/main/protocol.go#L71. If we have 2 policy function one calling with the remote and other with the local address, then i felt its not straight forward on handling different policy returned when both the functions are used.
I can change it if we are okay with second function overriding if both are set.

kmala: @pires what do you suggest as i don't see any other option apart from these?

pires: I'm thinking if anything, we'd split `PolicyFunc` into `DownstreamPolicyFunc` and `UpstreamPolicyFunc` with the same signature. My only _strong_ opinion is not wanting to break the API.

pires: @emersion got any thoughts on this? 🙏🏻

kmala: Yes, i understand that. Do we want to have one function override the other if both are present or do we want to pick a stricter policy of both(which i think is bit difficult to say which is stricter)

protocol.go
```diff
@@ -68,7 +68,7 @@ func (p *Listener) Accept() (net.Conn, error) {
 
 	proxyHeaderPolicy := USE
 	if p.Policy != nil {
-		proxyHeaderPolicy, err = p.Policy(conn.RemoteAddr())
+		proxyHeaderPolicy, err = p.Policy(conn.RemoteAddr(), conn.LocalAddr())
```
pires, line 71: I see now what the problem is. Let me think a little bit more about this. We'll find agreement.

kmala: sure 👍

kmala: @pires did you get a chance to look more into this?

emersion:
If we don't want to break the API, my take would be to add a new `ConnPolicy` callback:

- Only one of `Policy` and `ConnPolicy` can be specified, not both, to avoid having to pick between the two.
- `ConnPolicy` takes the `net.Conn` so that it can fetch any details it needs about the connection. Alternatively, it takes a `ConnPolicyOptions` struct which carry the IP addresses and can be extended later with more fields.

kmala:
> Only one of Policy and ConnPolicy can be specified, not both, to avoid having to pick between the two.

The issue is that since the Listener is exported https://github.com/pires/go-proxyproto/blob/main/protocol.go#L25 there is no proper way to enforce that only one of the policy/connpolicy can be specified.

We can just ignore Policy if ConnPolicy is specified in the code but that won't be obvious unless someone reads the code or comments.

emersion:
It should be possible to panic from `Accept` if the library user provides both. Not ideal, but still better than silently ignoring one of the callbacks IMHO.

kmala:
@emersion can you please check if the changes look good?

pires:
I am thinking the `ConnPolicy` should become the new API, since it seems to encapsulate what `Policy` is doing today, and we'd mark `Policy` as deprecated and to be removed in a future release. WDYT?

kmala:
> I am thinking the `ConnPolicy` should become the new API, since it seems to encapsulate what `Policy` is doing today, and we'd mark `Policy` as deprecated and to be removed in a future release. WDYT?

Update the field as deprecated.

kmala:
@pires @emersion can you please review when you get chance ?

pires (approved):

pires:
Thanks a ton for this contribution. We'll be releasing a new minor for introducing the new behavior and announce the deprecation of `PolicyFunc` 🖖🏻

PR #110, opened 2024-06-14
Protocol: Keep listener listening if we don't trust the upstream address

peteski22:
This PR is designed to prevent listeners being stopped when an error is returned, if the upstream connection address is not trusted (`ErrInvalidUpstream`). Instead, we continue to close the connection but now the `Accept` method has a `for` loop to continue looking for other connections to accept.

In using this library we discovered that the listener `Accept` method returning an error caused the listener to be closed and never reopened when trying to serve HTTP endpoints.

The change was based on [github.com/armon/go-proxyproto/](https://github.com/armon/go-proxyproto/blob/master/protocol.go#L70) which does something similar with the loop and checking for a particular type of error.

Notes: 

See: [`net/http/server.go => Serve(l net.Listener)`](https://cs.opensource.google/go/go/+/master:src/net/http/server.go;l=3300)

https://cs.opensource.google/go/go/+/master:src/net/http/server.go;l=3333-3351

```go
rw, err := l.Accept()
if err != nil {
	if srv.shuttingDown() {
		return ErrServerClosed
	}
	if ne, ok := err.(net.Error); ok && ne.Temporary() {
		if tempDelay == 0 {
			tempDelay = 5 * time.Millisecond
		} else {
			tempDelay *= 2
		}
		if max := 1 * time.Second; tempDelay > max {
			tempDelay = max
		}
		srv.logf("http: Accept error: %v; retrying in %v", err, tempDelay)
		time.Sleep(tempDelay)
		continue
	}
	return err
}
```

Above we end up returning the `err` at the end, which stops the `Serve` method (so prevents us listening).

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -2,6 +2,8 @@ package proxyproto
 
 import (
 	"bufio"
+	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sync"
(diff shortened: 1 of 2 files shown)


pires:
Can you, please, rebase?

peteski22:
> Can you, please, rebase?

@pires done now, thanks.

camaeel:
@pires any chance for merging this one?

pires:
Sorry for the delay. Can we have some tests proving the changes work as expected?

peteski22:
@pires no worries, I've added a test to show the listener stays open when experiencing the invalid upstream error but closes on others.

peteski22:
@pires there doesn't seem to be any issues now other than linting which this PR didn't introduce. Would you consider merging it? 🙏🏼

pires:
https://github.com/pires/go-proxyproto/pull/117/#issuecomment-2399563695

peteski22:
> Unfortunately, your PR history can't be rebase due to conflicts. I've fixed them but would like for you to re-author the commit so get proper recognition 🙏🏻 #117

Hi @pires thanks for the message, it's much appreciated. I've rebased everything from `upstream/main` now so I'm hoping it should look right and this PR can be merged now? 🤞🏼

pires:
Superb! Thanks a ton.

Closes #117

PR #114, opened 2024-08-10
policy: add REFUSE

drakkan:
in strict whitelist policies we want to refuse a connection from a not allowed upstream address whether the proxy header is set or not set.

Before this change if the upstream address is not allowed:

1) if the policy returns REJECT, the connection is allowed if no proxy
   header is sent
2) if the policy returns REQUIRE, the connection is allowed if a proxy
   header is set, even if the upstream address is not allowed to set it.

The new REFUSE policy can be returned for not allowed addresses so that the connection is always refused.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -51,6 +51,9 @@ const (
 	// Note: an example usage can be found in the SkipProxyHeaderForCIDR
 	// function.
 	SKIP
+	// REFUSE is the same as REJECT if a proxy header is set and the same as
+	// REQUIRE if a proxy header is not set.
+	REFUSE
 )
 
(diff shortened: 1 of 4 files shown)


drakkan:
I just noticed #110: if we join that PR, another way to reject a disallowed IP is to return an error from the policy function, while with this PR we can do `return proxyproto.REFUSE, nil` but in this case the connection errors out after reading the header instead of failing in Accept

drakkan:
closing, once #110 is merged we don't need a new policy we can just return an error for not allowed upstream ip addresses

PR #115, opened 2024-09-03
Fix: use v6 when source ip and dest ip version not equal

ghost:
When either the source ip or the destination ip is IPv6, use v6

### DIFF
--- a/header.go
+++ b/header.go
@@ -61,21 +61,23 @@ func HeaderProxyFromAddrs(version byte, sourceAddr, destAddr net.Addr) *Header {
 	}
 	switch sourceAddr := sourceAddr.(type) {
 	case *net.TCPAddr:
-		if _, ok := destAddr.(*net.TCPAddr); !ok {
+		destAddr, ok := destAddr.(*net.TCPAddr)
+		if !ok {
 			break
 		}
-		if len(sourceAddr.IP.To4()) == net.IPv4len {
+		if isIPv4(sourceAddr.IP) && isIPv4(destAddr.IP) {
 			h.TransportProtocol = TCPv4
-		} else if len(sourceAddr.IP) == net.IPv6len {
+		} else {
 			h.TransportProtocol = TCPv6
 		}
 	case *net.UDPAddr:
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -61,22 +61,24 @@ func HeaderProxyFromAddrs(version byte, sourceAddr, destAddr net.Addr) *Header {
...
 			break
 		}
-		if len(sourceAddr.IP.To4()) == net.IPv4len {
-			h.TransportProtocol = TCPv4
-		} else if len(sourceAddr.IP) == net.IPv6len {
+		if isIPv6(sourceAddr.IP) || isIPv6(destAddr.IP) {
 			h.TransportProtocol = TCPv6
+		} else if isIPv4(sourceAddr.IP) && isIPv4(destAddr.IP) {
+			h.TransportProtocol = TCPv4
 		}
```
AlexanderYastrebov, line 72:
```go
if isIPv4(sourceAddr.IP) && isIPv4(destAddr.IP) {
  h.TransportProtocol = TCPv4
} else {
  h.TransportProtocol = TCPv6
}
```

utils.go
```diff
@@ -0,0 +1,13 @@
...
+// https://www.haproxy.org/download/2.3/doc/proxy-protocol.txt
+package proxyproto
+
+import "net"
+
+func isIPv4(ip net.IP) bool {
+	return ip.To4() != nil
+}
+
+func isIPv6(ip net.IP) bool {
```
AlexanderYastrebov, line 11: This is not used and you can inline ip.To4() != nil and get rid of this file altogether.

utils.go
```diff
@@ -0,0 +1,13 @@
+// Package proxyproto implements Proxy Protocol (v1 and v2) parser and writer, as per specification:
+// https://www.haproxy.org/download/2.3/doc/proxy-protocol.txt
+package proxyproto
+
+import "net"
+
+func isIPv4(ip net.IP) bool {
+	return ip.To4() != nil
```
AlexanderYastrebov, line 8: This might not be correct as it also handles IPv6 mapped addresses. Maybe keep checking length like it was doe previously.

Issue #106, opened 2024-02-20
Suggestion: Dynamic Proxy Protocol Listener


134130:
- There can be requirements that server can listen and serve whenever proxy protocol is set or not for identifying real client's ip.
- I will make PR if many people needs it.

pires:
> It's a very useful feature. I'm waiting for your PR :)

Same!

dgl:
Implementing such a thing would like open a security hole, see the spec: https://www.haproxy.org/download/3.1/doc/proxy-protocol.txt

```
5. Security considerations

[...]

Implementers should be very careful about not trying to automatically detect
whether they have to decode the header or not, but rather they must only rely
on a configuration parameter. Indeed, if the opportunity is left to a normal
client to use the protocol, it will be able to hide its activities or make them
appear as coming from somewhere else. However, accepting the header only from a
number of known sources should be safe.
```

134130:
Yes, but it depends on the user's choice.

Restricting on the library level makes sense, but if the feature is useful for testing, it makes sense to support it, too. There are some cases like InsecureSkipTLSVerify for curl.

But the feature is not willing by folks, so there is no need to implement it immediately.

PR #116, opened 2024-09-17
protocol: avoid double buffering

mmatczuk:
Use smaller buffer size, and use buffer only to read the PROXY header.

Users may use they own buffers with they own buffer sizes and pools - connection should respect that.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -134,8 +134,13 @@ func (p *Listener) Addr() net.Addr {
 // NewConn is used to wrap a net.Conn that may be speaking
 // the proxy protocol into a proxyproto.Conn
 func NewConn(conn net.Conn, opts ...func(*Conn)) *Conn {
+	// For v1 the header length is at most 108 bytes.
+	// For v2 the header length is at most 52 bytes plus the length of the TLVs.
+	// We use 256 bytes to be safe.
+	const bufSize = 256
+
 	pConn := &Conn{
-		bufReader: bufio.NewReader(conn),
+		bufReader: bufio.NewReaderSize(conn, bufSize),
 		conn:      conn,
 	}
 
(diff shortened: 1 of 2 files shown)


mmatczuk:
I fixed the listing issues in the patch, and piggybacked patch to fix the ioutil usage.

pires:
Hey there! Thank you for your contribution. Would you be so kind as to check the tests failure? https://github.com/pires/go-proxyproto/actions/runs/10919900083/job/30323349143?pr=116#step:5:194

mmatczuk:
Sure, it's done.

pires:
I am paying attention to this but I need #110 to land first 🙏🏻 thanks for your patience.

pires:
Actually, I'm merging this now. Thanks a lot for your contribution.

protocol.go
```diff
@@ -134,8 +135,15 @@ func (p *Listener) Addr() net.Addr {
...
+	// For v1 the header length is at most 108 bytes.
+	// For v2 the header length is at most 52 bytes plus the length of the TLVs.
+	// We use 256 bytes to be safe.
+	const bufSize = 256
+	br := bufio.NewReaderSize(conn, bufSize)
+
 	pConn := &Conn{
-		bufReader: bufio.NewReader(conn),
+		bufReader: br,
+		reader:    io.MultiReader(br, conn),
```
AlexanderYastrebov, line 146: I got the idea but this does not look right: io.MultiReader will read br reader until EOF but it is a buffered reader that will read until connection returns EOF.

AlexanderYastrebov: I think it should be something like https://github.com/pires/go-proxyproto/pull/119

AlexanderYastrebov:
> Taking a wider perspective I think that it would be better to have the parsing APIs based on io.Reader and use ReadAll instead of Peek(). This is out of scope of this pr.

Yes, but then again one needs to keep a copy of read bytes in case there is no PROXY header to support https://github.com/pires/go-proxyproto/issues/106 usecase.

PR #119, opened 2024-10-10
Buffer only proxy header data

AlexanderYastrebov:
Reverts and re-implemnets incorrect https://github.com/pires/go-proxyproto/pull/116
which passed all connection data through the buffered reader.


pires:
@mmatczuk care to take a look at this?

mmatczuk:
I added my comments in https://github.com/pires/go-proxyproto/pull/116#issuecomment-2413378336.

mmatczuk:
Sent a followup patch https://github.com/pires/go-proxyproto/pull/120.

AlexanderYastrebov:
> Sent a followup patch #120.

It also has issues. Besides I fixed WriteTo instead of hand-rolled reads.

mmatczuk:
@AlexanderYastrebov 

> It also has issues.

PTAL at my comments in #120.

> Besides I fixed WriteTo instead of hand-rolled reads.

Sure, this is also fixed in #120

BTW adding the `WriteTo` to the interface if the underlying connection does not implement in may trick `io.Copy` with buffer from a buffer pool to falling back to plain io.Copy that allocates. Also the `WriteTo` only works on Linux IF the writer is UnixConn.

Cheers

pires:
Fixed in #148

PR #121, opened 2025-01-07
Fix Go Documentation of Default timeout

TheM1984:
With commit 7f1f93d the default timeout was changed from 200ms to 10s however the Go Documentation still shows the old default.

This commit fixes that, also I clarified in the protocol_test.go why the DefaultReadHeader is changed to 200ms.

No extra tests are needed since there is no Go code added, changed or removed.


pires (approved):
Nice catch, thank you.

PR #125, opened 2025-05-02
bump Go and golang.org/x/net

pires:
Closes #124

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/pires/go-proxyproto
 
-go 1.18
+go 1.24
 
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/pires/go-proxyproto
 
-go 1.18
```
jaimem88, line 3: This breaks backwards compatibility with modules built with older Go versions, for example I can't build my service using Go 1.23 https://gitlab.com/gitlab-renovate-forks/gitlab-pages/-/jobs/9927946165

PR #127, opened 2025-05-12
ci: use local Go toolchain and downgrade Go version

### LINKED ISSUE #126: Cannot builld using Go <= 1.23
This breaks backwards compatibility with modules built with older Go versions, for example I can't build my service using Go 1.23 https://gitlab.com/gitlab-renovate-forks/gitlab-pages/-/jobs/9927946165

_Originally posted by @jaimem88 in https://github.com/pires/go-proxyproto/pull/125#discussion_r2073173108_

gustavosbarreto:
In our GitHub Actions workflow we declare:

```yaml
strategy:
  matrix:
    go: ['1.23', '1.24']
```

However, the `1.23` job still ends up running Go `1.24`. By default, `actions/setup-go` leaves `GOTOOLCHAIN=auto`, causing go itself to fetch the version specified in `go.mod` instead instead of sticking to exactly `1.23`.

This led to a false positive in CI: it appeared as if the matrix was testing Go `1.23`, when in reality all jobs were using Go `1.24`.

To fix this, we need to force the use of the local Go toolchain in the CI test workflow to ensure the use of the locally installed Go version and avoid automatic downloads by `setup-go`.

Also downgrades the Go version in `go.mod` from `1.24` to `1.23` to ensure compatibility with environments that do not yet support Go `1.24`.

Fixes #126


pires (approved):
Thanks a ton!

pires:
Somehow, the linter is asking for `go mod tidy` :( WDYT?

gustavosbarreto:
I'll check this

gustavosbarreto:
> So it seems the sums for the deps change depending on the Go version being used, am I reading the `go mod tidy` correctly?

Yep, even if `go mod tidy` doesn't change `go.sum`, it still refreshes the module state internally. That helps tools like `golangci-lint`, which rely on `go list` under the hood, avoid context loading errors in clean or containerized environments.

gustavosbarreto:
> So it seems the sums for the deps change depending on the Go version being used, am I reading the `go mod tidy` correctly?

Yep, even if `go mod tidy` doesn't change `go.sum`, it still refreshes the module state internally. That helps tools like `golangci-lint`, which rely on `go list` under the hood, avoid context loading errors in clean or containerized environments.

pires:
@pkaeding a new release `v0.9.0` was made which includes updates to Go 1.24 but also tests 1.25, since 1.23 is EOL.

PR #148, opened 2026-02-04
protocol: don't buffer underlying conn reads after parsing header

### LINKED ISSUE #147: protocol: `io.MultiReader` stuck on `bufio.Reader.Read(b)`
`io.MultiReader` only advances to the next reader when the current reader returns `io.EOF`. A `bufio.Reader` wrapping an open `net.Conn` will not return EOF when its buffer is empty; it will instead read more data from the underlying connection to refill its buffer. This results in the `io.MultiReader` to get stuck on the `bufio.Reader` for the entire lifetime of the connection, which effectively makes the direct `conn.Read(b)`  unreachable dead code.

For more info, see #142 #123 #120 #119

pires:
Fixes #120 
Fixes #147
Refs #142

Please, critique @clementnuss @emersion.

### DIFF
--- a/protocol_test.go
+++ b/protocol_test.go
@@ -2458,6 +2458,91 @@ func TestConnReadHandlesChunkedPayload(t *testing.T) {
 	}
 }
 
+func TestReadUsesConnWhenBufReaderNil(t *testing.T) {
+	serverConn, clientConn := net.Pipe()
+	t.Cleanup(func() {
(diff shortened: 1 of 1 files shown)


emersion:
Oh, actually, I think we need to add an extra check to the `WriteTo` impl. As-is, it may panic accessing an nil `p.bufReader`.

PR #144, opened 2026-02-03
Bump maximum header size to 4096 bytes

emersion:
PP2_SUBTYPE_SSL_CLIENT_CERT is typically around 1 or 2 kilobytes.

PR #150, opened 2026-02-04
protocol: safely resize header buffer based on length field

pires:
Fixes #123 
Refs #144 

WDYT @emersion @bollenberger?

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -2,6 +2,8 @@ package proxyproto
 
 import (
 	"bufio"
+	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


protocol.go
```diff
@@ -300,6 +303,24 @@ func (p *Conn) readHeader() error {
...
+	// does not consume bytes; it avoids allocations unless the header looks
+	// well-formed enough to trust its length field.
+	// We need 16 bytes: 12 (Signature) + 1 (Version/Command) + 1 (Transport) + 2 (Length).
+	if p.bufReader != nil {
+		peeked, err := p.bufReader.Peek(16)
+		if err == nil {
+			if totalSize, ok := v2PreflightHeaderSize(peeked); ok {
+				// If the header is larger than our current buffer, wrap the existing
+				// reader in a larger one. This preserves any bytes already buffered.
+				if totalSize > p.bufReader.Size() {
```
emersion: This can result in a DoS if the client sends 65535 as the length.

PR #167, opened 2026-03-18
v1: accept plain IPv4 addresses in TCP6 headers

cmarker-gl:
## Problem

When proxying SSH connections through a chain involving nginx's OSS stream module, PROXY protocol v1 headers with mixed address families are emitted and currently rejected by this library.

The chain looks like:

```
IPv6 client → Cloudflare Spectrum (PROXY v2) → nginx stream (decodes v2, re-encodes as v1) → application (go-proxyproto)
```

nginx's OSS stream module (`proxy_protocol on`) always re-encodes as PROXY protocol v1. When the original client is IPv6 but the backend is IPv4, nginx emits:

```
PROXY TCP6 2001:db8::1 192.0.2.1 51512 22\r\n
```

This is technically outside the strict spec (which says both addresses in a `TCP6` header should be in IPv6 format), but it is the real-world output of a widely deployed proxy and there is no workaround available in nginx OSS.

go-proxyproto was returning `proxyproto: invalid address` for these headers, causing the connection to fail. IPv4-only connections worked fine.

## Root Cause

`parseV1IPAddress` in `v1.go` handled `TCPv6` by accepting addresses where `addr.Is6()` or `addr.Is4In6()` (i.e. `::ffff:x.x.x.x` notation). A plain IPv4 address like `192.0.2.1` parses as `addr.Is4()` -- neither condition matched, so it fell through to `ErrInvalidAddress`.

## Fix

When parsing a `TCPv6` header and the address is a plain IPv4, promote it to a 16-byte IPv4-mapped-IPv6 representation via `netip.AddrFrom16(addr.As16())`. This is consistent with how `net.IP.To16()` works elsewhere in the library and produces a valid `net.IP` that callers can use normally.

Note: this is a lossy conversion -- round-trip serialization will render the address as `::ffff:x.x.x.x` rather than the original `x.x.x.x`.

## Tests

- Added `"TCP6 with invalid address"` test case (`PROXY TCP6 not-an-ip ::1 1234 5678`)
- Added three new valid parse cases covering mixed address families in TCP6 headers:
  - `PROXY TCP6 192.0.2.1 192.0.2.2 1234 5678` -- IPv4 src, IPv4 dst
  - `PROXY TCP6 2001:db8::1 192.0.2.1 51512 22` -- IPv6 src, IPv4 dst (the nginx mixed case)
  - `PROXY TCP6 192.0.2.1 2001:db8::1 51512 22` -- IPv4 src, IPv6 dst
- All mixed-address test cases use `skipWrite: true` since round-trip serialization changes the address representation
- Removed duplicate test fixtures that were identical to existing `fixtureTCP4V1` and `fixtureTCP6V1`

## Notes

- PROXY protocol v2 is **not affected** -- v2 is binary and reads fixed-width 16-byte address fields directly, so mixed address families cannot arise in the same way
- No API changes; downstream consumers need only bump their dependency to pick up the fix
- This code was developed in part with [GitLab Duo](https://about.gitlab.com/gitlab-duo/)

### DIFF
--- a/v1.go
+++ b/v1.go
@@ -237,12 +237,20 @@ func parseV1IPAddress(protocol AddressFamilyAndProtocol, addrStr string) (net.IP
 			return net.IP(addr.AsSlice()), nil
 		}
 	case TCPv6:
+		// Some proxies (notably nginx OSS stream module) emit plain IPv4
+		// addresses in TCP6 headers when the backend is IPv4 but the client
+		// is IPv6. Promote to IPv4-mapped IPv6 for interoperability.
+		//
+		// This is an intentional departure from the PROXY protocol v1 spec,
+		// which states that addresses in a TCP6 line must be in IPv6 format.
 		if addr.Is6() || addr.Is4In6() {
 			return net.IP(addr.AsSlice()), nil
 		}
+		// ATTENTION: this is a lossy conversion — round-trip serialization will
+		// render the address as "::ffff:x.x.x.x" rather than the original "x.x.x.x".
 		if addr.Is4() {
-			a16 := addr.As16()
-			return net.IP(a16[:]), nil
+			mapped := netip.AddrFrom16(addr.As16())
+			return net.IP(mapped.AsSlice()), nil
 		}
 	}
 
--- a/v1.go
+++ b/v1.go
@@ -240,6 +240,10 @@ func parseV1IPAddress(protocol AddressFamilyAndProtocol, addrStr string) (net.IP
 		if addr.Is6() || addr.Is4In6() {
 			return net.IP(addr.AsSlice()), nil
 		}
+		if addr.Is4() {
+			a16 := addr.As16()
+			return net.IP(a16[:]), nil
+		}
 	}
 
 	return nil, ErrInvalidAddress
--- a/v1_test.go
+++ b/v1_test.go
@@ -29,9 +29,9 @@ var (
 	fixtureUnknown              = "PROXY UNKNOWN" + crlf
 	fixtureUnknownWithAddresses = "PROXY UNKNOWN " + IPv4AddressesAndInvalidPorts + crlf
 
-	fixtureTCP4Simple         = "PROXY TCP4 1.2.3.4 5.6.7.8 12345 22" + crlf
+	fixtureTCP6IPv4SrcIPv4Dst = "PROXY TCP6 192.0.2.1 192.0.2.2 1234 5678" + crlf
 	fixtureTCP6IPv6SrcIPv4Dst = "PROXY TCP6 2001:db8::1 192.0.2.1 51512 22" + crlf
-	fixtureTCP6Loopback       = "PROXY TCP6 ::1 ::1 1234 5678" + crlf
+	fixtureTCP6IPv4SrcIPv6Dst = "PROXY TCP6 192.0.2.1 2001:db8::1 51512 22" + crlf
 )
 
 var invalidParseV1Tests = []struct {
@@ -87,6 +87,11 @@ var invalidParseV1Tests = []struct {
 	{
 		desc:          "TCP4 with IPv4 mapped addresses",
 		reader:        newBufioReader([]byte("PROXY TCP4 " + IPv4In6AddressesAndPorts + crlf)),
+		expectedError: ErrInvalidAddress,
+	},
+	{
+		desc:          "TCP6 with invalid address",
+		reader:        newBufioReader([]byte("PROXY TCP6 not-an-ip ::1 1234 5678" + crlf)),
 		expectedError: ErrInvalidAddress,
 	},
 	{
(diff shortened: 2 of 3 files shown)


cmarker-gl:
@pires can you take a look at this PR and let me know what you think?

pires:
Hi @cmarker-gl sorry but I was AFK past week. I now have some work to attend to, and will come back to you soon.

v1_test.go
```diff
@@ -177,6 +176,40 @@ var validParseAndWriteV1Tests = []struct {
 			DestinationAddr:   nil,
 		},
 	},
+	{
+		desc:   "TCP4 simple",
+		reader: bufio.NewReader(strings.NewReader(fixtureTCP4Simple)),
+		expectedHeader: &Header{
```
pires, line 187: Same as `fixtureTCP4V1`.

v1_test.go
```diff
@@ -177,6 +176,40 @@ var validParseAndWriteV1Tests = []struct {
...
+			Version:           1,
+			Command:           PROXY,
+			TransportProtocol: TCPv6,
+			SourceAddr:        &net.TCPAddr{IP: net.ParseIP("2001:db8::1").To16(), Port: 51512},
+			DestinationAddr:   &net.TCPAddr{IP: net.ParseIP("192.0.2.1").To16(), Port: 22},
+		},
+		skipWrite: true,
+	},
+	{
+		desc:   "TCP6 loopback",
```
pires: Same as `fixtureTCP6V1`.

v1_test.go
```diff
@@ -75,11 +79,6 @@ var invalidParseV1Tests = []struct {
 		reader:        newBufioReader([]byte("PROXY TCP4 invalid invalid 65533 65533" + crlf)),
 		expectedError: ErrInvalidAddress,
 	},
-	{
```
pires, line 78: I wouldn't delete this and instead rework it to include a TCP6 w/ invalid address, such as:

```go
{
    desc:          "TCP6 with invalid address",
    reader:        newBufioReader([]byte("PROXY TCP6 not-an-ip ::1 1234 5678" + crlf)),
    expectedError: ErrInvalidAddress,
},
```

v1.go
```diff
@@ -240,6 +240,10 @@ func parseV1IPAddress(protocol AddressFamilyAndProtocol, addrStr string) (net.IP
 		if addr.Is6() || addr.Is4In6() {
 			return net.IP(addr.AsSlice()), nil
 		}
+		if addr.Is4() {
+			a16 := addr.As16()
+			return net.IP(a16[:]), nil
```
pires: In order to keep pattern consistent (see line 241), I would change this to something like:

```go
mapped := netip.AddrFrom16(addr.As16())
return net.IP(mapped.AsSlice()), nil
```

v1.go
```diff
@@ -240,6 +240,10 @@ func parseV1IPAddress(protocol AddressFamilyAndProtocol, addrStr string) (net.IP
 		if addr.Is6() || addr.Is4In6() {
```
pires, line 246:
```suggestion
  		// Some proxies (notably nginx OSS stream module) emit plain IPv4
  		// addresses in TCP6 headers when the backend is IPv4 but the client
  		// is IPv6. Promote to IPv4-mapped IPv6 for interoperability.
  		//
 		// This is an intentional departure from the PROXY protocol v1 spec,
  		// which states that addresses in a TCP6 line must be in IPv6 format.
		if addr.Is6() || addr.Is4In6() {
```

v1.go
```diff
@@ -240,6 +240,10 @@ func parseV1IPAddress(protocol AddressFamilyAndProtocol, addrStr string) (net.IP
...
 			return net.IP(addr.AsSlice()), nil
 		}
+		if addr.Is4() {
```
pires, line 251:
```suggestion
		// ATTENTION: this is a lossy conversion — round-trip serialization will
		// render the address as "::ffff:x.x.x.x" rather than the original "x.x.x.x".
		if addr.Is4() {
```

pires (commented):
First of all, thank you for your contribution. And then thank you for your patience, I have been immensely busy.

Please, fix the linter complain(s) and add tests for the following scenarios:
* IPv4 src, IPv4 dst in `TCP6: PROXY TCP6 192.0.2.1 192.0.2.2 1234 5678`
* IPv4 src, IPv6 dst in TCP6: the reversed mixed case

cmarker-gl:
Thanks for the review! I've pushed two commits addressing your feedback:

**Linter fix** (`41f2c96`): Added a `//nolint:gosec` directive for the G706 finding in `examples/httpserver/httpserver.go` since logging the remote address is intentional in that example.

**Review feedback** (`aa3ddcd`): Applied all your suggestions (comments, `netip.AddrFrom16` refactor, removed duplicate fixtures, reworked the invalid-address test for TCP6, added the two mixed-address test scenarios). Updated the PR description as well.

v1_test.go
cmarker-gl: Good call!

pires (approved):
Thanks a ton for your contribution 🙇🏻

cmarker-gl:
Appreciate the merge! Just need a version tag and release pushed to wrap this up. Let me know if you need anything from my end.

pires:
🙇🏻‍♂️ https://github.com/pires/go-proxyproto/releases/tag/v0.12.0

PR #168, opened 2026-04-17
header: use bytes.Reader instead of bytes.Buffer

rbqvq:
Use bytes.Reader instead of bytes.Buffer.

### DIFF
--- a/header.go
+++ b/header.go
@@ -199,7 +199,7 @@ func (header *Header) WriteTo(w io.Writer) (int64, error) {
 		return 0, err
 	}
 
-	return bytes.NewBuffer(buf).WriteTo(w)
+	return bytes.NewReader(buf).WriteTo(w)
 }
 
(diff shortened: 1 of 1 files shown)


pires:
Please, share your rationale.

rbqvq:
`bytes.Reader` is designed for read-only data, whereas `bytes.Buffer` is a growable variable-sized buffer.

For better semantic clarity and lower overhead, I recommend using `bytes.Reader` instead of `bytes.Buffer`.

pires:
Thank you! Let's wait for #173 given this is a single commit that can easily be rebased and pushed just by yourself 🙇🏻‍♂️

pires:
Please, rebase and push. Thank you!

pires (approved):

rbqvq:
Thank you, Wait the next tag version 🎉

PR #169, opened 2026-04-17
Improve IP address handling for v1 and version detection

rbqvq:
The standard `net.IP` implementation inconsistently represents IPv4-mapped IPv6 addresses; its `String()` method returns an IPv4 string for mapped addresses, which causes formatting issues in v1 headers where strict version representation is required. (#167)

This patch migrates IP state management to `netip.Addr` to ensure that `String()` outputs and byte representations strictly follow the expected protocol format.

Additionally, this fix addresses a bug in the header construction logic. 

The previous implementation only verified the source IP version, leading to incorrect protocol family assignments when source and destination address versions mismatched.

The logic now ensures that TCPv6/UDPv6 is selected if either address is IPv6, with IPv4 addresses correctly formatted as IPv4-mapped IPv6 addresses when necessary.

- Migrate from `net.IP` to `netip.Addr` for robust IP handling
- Update version detection to fallback to IPv6 if any address is IPv6
- Ensure strict 4-byte or 16-byte alignment in v1 header output

Migrate from https://github.com/pires/go-proxyproto/pull/115

### DIFF
--- a/header.go
+++ b/header.go
@@ -78,21 +78,25 @@ func HeaderProxyFromAddrs(version byte, sourceAddr, destAddr net.Addr) *Header {
 	}
 	switch sourceAddr := sourceAddr.(type) {
 	case *net.TCPAddr:
-		if _, ok := destAddr.(*net.TCPAddr); !ok {
+		destAddr, ok := destAddr.(*net.TCPAddr)
+		if !ok {
 			break
 		}
-		if len(sourceAddr.IP.To4()) == net.IPv4len {
+		switch {
+		case sourceAddr.IP.To4() != nil && destAddr.IP.To4() != nil:
 			h.TransportProtocol = TCPv4
-		} else if len(sourceAddr.IP) == net.IPv6len {
+		case sourceAddr.IP.To16() != nil && destAddr.IP.To16() != nil:
 			h.TransportProtocol = TCPv6
 		}
 	case *net.UDPAddr:
-		if _, ok := destAddr.(*net.UDPAddr); !ok {
+		destAddr, ok := destAddr.(*net.UDPAddr)
+		if !ok {
 			break
 		}
-		if len(sourceAddr.IP.To4()) == net.IPv4len {
+		switch {
+		case sourceAddr.IP.To4() != nil && destAddr.IP.To4() != nil:
 			h.TransportProtocol = UDPv4
-		} else if len(sourceAddr.IP) == net.IPv6len {
+		case sourceAddr.IP.To16() != nil && destAddr.IP.To16() != nil:
 			h.TransportProtocol = UDPv6
 		}
 	case *net.UnixAddr:
--- a/header.go
+++ b/header.go
@@ -85,21 +85,25 @@ func HeaderProxyFromAddrs(version byte, sourceAddr, destAddr net.Addr) *Header {
 	}
 	switch sourceAddr := sourceAddr.(type) {
 	case *net.TCPAddr:
-		if _, ok := destAddr.(*net.TCPAddr); !ok {
+		destAddr, ok := destAddr.(*net.TCPAddr)
+		if !ok {
 			break
 		}
-		if len(sourceAddr.IP.To4()) == net.IPv4len {
+		switch {
+		case sourceAddr.IP.To4() != nil && destAddr.IP.To4() != nil:
 			h.TransportProtocol = TCPv4
-		} else if len(sourceAddr.IP) == net.IPv6len {
+		case sourceAddr.IP.To16() != nil && destAddr.IP.To16() != nil:
 			h.TransportProtocol = TCPv6
 		}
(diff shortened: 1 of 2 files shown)


rbqvq:
@pires may you take look this?

pires:
@rbqvq maybe you forgot to push the fix for the issue breaking the tests?

v1.go
```diff
@@ -184,16 +184,16 @@ func (header *Header) formatVersion1() ([]byte, error) {
...
 	case TCPv4:
-		sourceIP = sourceIP.To4()
-		destIP = destIP.To4()
+		sourceIP, _ = netip.AddrFromSlice(sourceAddr.IP.To4())
+		destIP, _ = netip.AddrFromSlice(destAddr.IP.To4())
 	case TCPv6:
-		sourceIP = sourceIP.To16()
-		destIP = destIP.To16()
+		sourceIP, _ = netip.AddrFromSlice(sourceAddr.IP.To4())
+		destIP, _ = netip.AddrFromSlice(destAddr.IP.To4())
```
pires:
```suggestion
		destIP, _ = netip.AddrFromSlice(destAddr.IP.To16())
```

v1.go
```diff
@@ -184,16 +184,16 @@ func (header *Header) formatVersion1() ([]byte, error) {
...
 	switch header.TransportProtocol {
 	case TCPv4:
-		sourceIP = sourceIP.To4()
-		destIP = destIP.To4()
+		sourceIP, _ = netip.AddrFromSlice(sourceAddr.IP.To4())
+		destIP, _ = netip.AddrFromSlice(destAddr.IP.To4())
 	case TCPv6:
-		sourceIP = sourceIP.To16()
-		destIP = destIP.To16()
+		sourceIP, _ = netip.AddrFromSlice(sourceAddr.IP.To4())
```
pires:
```suggestion
		// Use netip instead net.IP so to guarantee Is6() address, ie a v4-mapped IP in a TCP6 header
		// serializes as ::ffff:1.2.3.4 instead of net.IP.String()'s collapsed 1.2.3.4.
		sourceIP, _ = netip.AddrFromSlice(sourceAddr.IP.To16())
```

v1.go
```diff
@@ -184,16 +184,16 @@ func (header *Header) formatVersion1() ([]byte, error) {
...
+		destIP, _ = netip.AddrFromSlice(destAddr.IP.To4())
 	}
-	if sourceIP == nil || destIP == nil {
+	if !sourceIP.IsValid() || !destIP.IsValid() {
```
pires: Discarding the `ok` from `AddrFromSlice` and relying on `IsValid()` is acceptable (equivalent), but using `ok` directly reads clearer IMO.

header.go
```diff
@@ -78,21 +78,25 @@ func HeaderProxyFromAddrs(version byte, sourceAddr, destAddr net.Addr) *Header {
...
+		destAddr, ok := destAddr.(*net.TCPAddr)
+		if !ok {
 			break
 		}
-		if len(sourceAddr.IP.To4()) == net.IPv4len {
+		switch {
+		case sourceAddr.IP.To4() != nil && destAddr.IP.To4() != nil:
 			h.TransportProtocol = TCPv4
-		} else if len(sourceAddr.IP) == net.IPv6len {
+		case sourceAddr.IP.To16() != nil && destAddr.IP.To16() != nil:
```
pires, line 95: IIUC there's an imprecision here. `To16()` is non-nil for IPv4 too (it returns the 16-byte form of any valid 4 or 16-byte IP). So the v6 case really means "both are valid IPs and not both-v4" (the v4 case is tested first). This has the following consequences:
- For a mismatched-family pair (source v4 / dest v6), it now classifies TCPv6 (old code: TCPv4, from source). Mismatched families are nonsensical input either way, but it's a silent behavior change, and arguably `UNSPEC` (best-effort "leave unspecified", per the doc comment) would be the safer result.
- Clearer would be to compute each end's family once and require equality, or assert genuine v6 (`To4() == nil`) on both ends.

pires (changes requested):

header.go
rbqvq: In some cases, such as when we need to read an address from Nginx OSS without rewriting and directly forwarding it, Nginx might send 4in6 address.

https://github.com/pires/go-proxyproto/pull/167

We won't send non-standard TCPv6 header like Nginx OSS, we send 4in6 addresses.

Previously this was UNSPEC, but now we specify it as IPv6, which I think is the expected result.

I personally think this is acceptable; the 4in6 address in IPv6 standard is used for this purpose.

Also, net.ParseIP and net.IPv4 always return 16 bytes unless `To4` is called manually.

I'd like to know your opinion.

pires:
@rbqvq please, review #173 which includes your changes and some of my own.

rbqvq:
LGTM

PR #173, opened 2026-06-30
Improve IP family handling

pires:
Fixes #169

### v1

- `default: return nil, ErrInvalidAddress` in the format switch

### Tests

- Dropped all four `skipWrite` flags
- Added byte-level wire assertion (the real regression lock)
- Added mixed-family case (`TCPv4SrcIPv6Dst` → `TCPv6`)


header.go
rbqvq, line 128: Should we check destAddt.Net are same with sourceAddr.Net?

pires: Sure, it's worth defending.

pires: Done and pushed.

rbqvq (approved):
LGTM

PR #178, opened 2026-07-03
BREAKING: from spec 2.3 to 3.4

### LINKED ISSUE #79: PacketConn interface
is it possible to add support for the udp net.PacketConn interface?

pires:
@emersion would be great to get your eyes on this massive PR 😅 as it will become a breaking change minor release - I really don't want to do a v2 Go module 😓 

## Summary

Aligns the library with PROXY protocol spec v3.4 (as of 2026/04/27).

### Highlights:

**1. v1/v2 parsing strictness** (`v1.go`, `v2.go`, `header.go`, `addr_proto.go`)
- v1: full `PROXY` token validated (rejects `PROXYjunk…`); IPv6 zone IDs rejected; plain IPv4 in `TCP6` lines rejected by default (nginx interop via new `V1AcceptIPv4InTCP6` var); ports parsed with no leading zeros; port range validated on the format side too.
- v2: family/transport whitelist now applies to LOCAL commands (undefined bytes like `0x99` rejected per §2.2); LOCAL with a valid address block decodes addresses/TLVs like PROXY and round-trips; LOCAL with a non-fitting declared length is skipped and normalized to UNSPEC so it stays serializable.
- New `MaxV2HeaderSize` var (default 4096) bounds the declared v2 length, overridable for large TLVs (e.g. `PP2_SUBTYPE_SSL_CLIENT_CERT`).
- `formatVersion2` folds TLV length into the declared length for Unix families (previously emitted TLVs past it).
- `Header.EqualsTo` no longer panics on nil addresses; `Conn.RemoteAddr`/`LocalAddr` fall back to the underlying conn when the header carries nil addresses.
- New `fuzz_test.go`: exhaustive 256-transport-byte sweeps for both commands, LOCAL decode matrix, byte-for-byte round-trip stability, `FuzzRead` parse→format→reparse invariants.

**2. Policy rework** (`policy.go`, `protocol.go`)
- New `DefaultPolicy` package var (`REQUIRE`), applied by `NewConn` and `Listener.Accept` when no policy is set — the spec forbids guessing whether header is present.
- `TrustProxyHeaderFrom` is now strict: trusted peers → REQUIRE, untrusted peers → dropped by `Accept` (previously USE/REJECT, which let untrusted headerless clients through as raw connections). New CIDR-capable sibling `TrustProxyHeaderFromRanges`.
- New `PolicyFromRanges(ranges, matched, unmatched)` + `MustPolicyFromRanges` combinator; the whole whitelist family (8 funcs) is deprecated and delegates to it.
- All built-in policies wrap address-classification failures in `ErrInvalidUpstream` so one unclassifiable peer (e.g. a Unix socket address can't stop an accept loop; contract documented on `PolicyFunc`/`ConnPolicyFunc`/`ErrInvalidUpstream`.
- `examples/httpserver` and the http2 helper tests opt into USE explicitly (they accept direct connections).

**3. TLV catch-up** (`tlv.go`, `tlvparse/`)
- `Registered()` includes the 3.4 types (`PP2_SUBTYPE_SSL_GROUP`, `PP2_SUBTYPE_SSL_SIG_SCHEME`, `PP2_SUBTYPE_SSL_CLIENT_CERT`).
- New `PP2SSL` accessors: `SSLSigAlg`, `SSLKeyAlg`, `SSLGroup`, `SSLSigScheme`.
- Coverage for malformed sub-TLVs and marshal errors; `tlvparse/test.go` renamed to `helpers_test.go`.

**4. UDP support** (`udp.go`)
- Spec 3.4 clarifies UDP transport: header and payload share one datagram, parsed independently per datagram. `Listener`/`Conn` are stream-oriented, so this adds `ParseUDPDatagram` / `Header.FormatUDPDatagram` for use with any `net.PacketConn`, plus godoc examples with `// Output` assertions.
- Deliberately **no** `net.PacketConn` implementation in the library: replying to the header-carried client address needs a client→proxy flow table whose bounds/expiry/spoofing exposure are application policy. `examples/udppacketconn` is a runnable sketch of that wrapper pattern (including LOCAL health-check handling); `examples/udpserver`/`udpclient` cover the plain per-datagram paths. Fixes #79


emersion:
Sorry, I don't have time to review such a huge PR.

pires:
Sorry I didn't give it enough time, mate! Meanwhile, tagged v0.15.0 and listed all changes, starting with the breaking ones. Hope this lands well in your codebase, if you still rely on it 🙏🏻

PR #182, opened 2026-07-30
fix: WriteTo empty buffer handling

### LINKED ISSUE #181: Read into empty buffer in Conn.WriteTo
In `Conn.WriteTo` if this condition is not true`p.bufReader == nil` but `p.bufReader.Buffered()` is zero there will be a read into empty buffer followed by zero byte write, which some systems may not expect:

https://github.com/pires/go-proxyproto/blob/b9877208baf4fffe3e2d267ae9b5b3ddb5f86968/protocol.go#L532

Same case is covered in `Conn.Read`:

https://github.com/pires/go-proxyproto/blob/b9877208baf4fffe3e2d267ae9b5b3ddb5f86968/protocol.go#L279

`github.com/pires/go-proxyproto v0.15.0`

mymmrac:
Added check for buffer size and cleaning of the buffer after reading.

This PR closes #181

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -524,15 +524,26 @@ func (p *Conn) WriteTo(w io.Writer) (int64, error) {
 		return 0, err
 	}
 
-	// If the buffer has been drained (or cleared), copy directly from conn.
+	// If the buffer has been cleared, copy directly from conn.
 	if p.bufReader == nil {
 		return io.Copy(w, p.conn)
 	}
 
+	// If the buffer has been drained, copy directly from conn.
+	if p.bufReader.Buffered() == 0 {
+		// Garbage collect the buffer.
+		p.bufReader = nil
+
+		return io.Copy(w, p.conn)
+	}
+
 	b := make([]byte, p.bufReader.Buffered())
-	if _, err := p.bufReader.Read(b); err != nil {
+	if _, err := io.ReadFull(p.bufReader, b); err != nil {
 		return 0, err // this should never happen as we read buffered data.
 	}
+
+	// Garbage collect the buffer.
+	p.bufReader = nil
 
 	var n int64
 	{
--- a/protocol.go
+++ b/protocol.go
@@ -532,8 +532,8 @@ func (p *Conn) WriteTo(w io.Writer) (int64, error) {
 	// If the buffer has been drained, copy directly from conn.
 	if p.bufReader.Buffered() == 0 {
 		// Check for pending errors.
-		if _, err := p.bufReader.Read(nil); err != nil {
-			return 0, err
+		if n, err := p.bufReader.Read(nil); err != nil {
+			return int64(n), err
 		}
 
 		// Garbage collect the buffer.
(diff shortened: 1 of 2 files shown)


protocol.go
```diff
@@ -524,16 +524,27 @@ func (p *Conn) WriteTo(w io.Writer) (int64, error) {
...
 	}
 
-	// If the buffer has been drained (or cleared), copy directly from conn.
+	// If the buffer has been cleared, copy directly from conn.
 	if p.bufReader == nil {
 		return io.Copy(w, p.conn)
 	}
 
+	// If the buffer has been drained, copy directly from conn.
+	if p.bufReader.Buffered() == 0 {
```
pires, line 533: IIUC `Buffered() == 0` only means no bytes remain; the reader can still hold a pending error from an underlying `Read` that returned (`n > 0, err`). Clearing `bufReader` and copying directly from `conn` discards that error.

pires (commented):
Thank you for your contribution.

IMO we should take the buffer-clearing optimization but do not discard an empty `bufio.Reader` without accounting for its pending error state. Also, and pretty please, add tests that assert both memory cleanup and preservation of underlying read errors.

protocol.go
mymmrac: Thanks for the comment, will try to find time to fix it over the weekends

mymmrac: @pires I added check for pending errors and tests that cover this case

mymmrac:
@pires if PR is okay, and you don't have further comments, please merge as I don't have permissions, thanks

pires:
I am looking into it with the time I have available, thank you for the ping.

mymmrac:
Of course, take your time, totally understand

protocol.go
```diff
@@ -524,16 +524,32 @@ func (p *Conn) WriteTo(w io.Writer) (int64, error) {
...
+		// Check for pending errors.
+		if _, err := p.bufReader.Read(nil); err != nil {
+			return 0, err
```
pires: Should we return any bytes read instead of `0`?

pires (commented):
Almost there, one minor question though.

protocol.go
mymmrac: When reading into `nil` (or empty buffer)  from `bufio.Reader` it always returns 0, so it's safe to return 0, but if you want, I can change it to be more explicit
```
func (b *Reader) Read(p []byte) (n int, err error) {
	n = len(p)
	if n == 0 {
		if b.Buffered() > 0 {
			return 0, nil
		}
		return 0, b.readErr()
	}
	// ...
```

pires: If it's always zero, `if read, err := p.bufReader.Read(nil); err != nil { return read, err }` would cut it, right?

mymmrac: Yep, will update

mymmrac: Hmm, I updated it, but then read this from `io.WriteTo` doc:
```
// WriteTo writes data to w until there's no more data to write or
// when an error occurs. The return value n is the number of bytes
// written. Any error encountered during the write is also returned.
```

So `n` is the number of bytes written, not number of bytes read, I suppose it should be 0 in case of read errors, no? What do you think?

pires: Oh, you're right, I'm sorry about that.

pires: Please, drop the last commit and I'll merge.

mymmrac: Reverted it

pires (approved):

